PR #10, opened 2019-05-04
New module stubs

pospi:
- Updated repo structure to make room for multiple DNAs & shared libraries, stubbed out a new DNA to proof 2 of them running in tandem
- Simplified task runner (now only 1 set of NPM commands)
- Cleaned up repository structure to play nicely with various editor expectations (VSCode doesn't like multiple Rust workspaces)
- Updated Holochain to latest version (`v0.0.14-alpha1`)
- Fleshed out contributor docs for performing common scaffolding tasks


pospi:
@sqykly if you're still running on Windows, you might have more luck setting up Holochain via the (recommended) Nix method, rather than installing via Rust. YMMV, but if this is the case it would be cool to:

- update `scripts/postinstall.sh` to detect whether Nix is present and install with that method rather than via `cargo` if it is found
- add instructions to the end of `CONTRIBUTORS.md` on how to update Holochain when Nix is managing it

You may want to ping the Holochain devs if you still don't have things set up and Windows is presenting a problem; or potentially run everything inside a VirtualBox VM. Hope it's painless!

pospi:
Oh, and you can confirm you have a working toolchain if you can do:

- `npm run build` successfully
- `npm run dht` successfully, once you have completed a build. Note the agent keyfile has an empty passphrase. You should see console output from both zomes' registration hooks after it loads.

PR #25, opened 2019-06-06
update field aliases to NewType pattern for strict type checking against input fields

pospi:
This will break some of the event / fulfillment test code, but @sqykly will likely replace that with his work anyway.


pospi:
Some open questions about the ergonomics of doing this- see https://chat.holochain.org/appsup/pl/c1dtckxyh3fpbx61gw8obzuqur

Issue #19, opened 2019-06-03
Implement bidirectional links between Commitment.satisfies & Intent.satisfiedBy


pospi:
This is a good first relationship to implement as it deals with same-DNA field links. Implementation requires:

- CREATE methods for `Commitment` & `Intent` in the planning DNA
    - Ability to create `Satisfaction` records as sub-records of a new `Commitment`
    - Ability to create `Satisfaction` records as sub-records of a new `Intent`
- CREATE method for `Satisfaction`
- UPDATE methods for `Commitment` & `Intent` (non-relational fields only)
- UPDATE method for `Satisfaction`
- DELETE method for `Satisfaction`
- DELETE methods for `Commitment` & `Intent`

I don't think there is any need to implement an ability to modify `Satisfaction` records via manipulation of `Commitment` or `Intent`- the client needs to be responsible for tracking which records to CREATE / DELETE anyway, so at that point they might as well call UPDATE / DELETE on `Satisfaction` directly. I do still think it's a nice convenience thing to have an ability to author them at `Commitment` / `Intent` logging time but am open to having my mind changed on that. I thought if nothing else it would be a good way to encourage splitting abstractions out.

We also need to define a naming convention for our zome API methods going forward, part of this issue is deciding on those naming conventions.

*(BTW: if you think we need to define acceptance criteria for this and other issues I'm about to log, please let me know; otherwise happy for those tests to emerge as makes sense.)*

fosterlynn:
>I don't think there is any need to implement an ability to modify Satisfaction records via manipulation of Commitment or Intent

Agree.

>I do still think it's a nice convenience thing to have an ability to author them at Commitment / Intent logging time but am open to having my mind changed on that.

Up to you guys of course, and I'm kind of like you @pospi , it would be nice, but not essential.  A couple thoughts.  

One step farther, another thing that is part of committing to an intent is that if the commitment completely satisfies the intent, then the intent can be marked finished=true.  

Another possible convenience, maybe going too far, is that probably most of the time, the satisfiedQuantity will be the same as the commitment quantity, so if not sent in as a parameter, that could be just assumed if you have the intent that was satisfied.  Same with fulfilledQuantity and the event quantity.

Not recommending in any direction, just providing info.

pospi:
I see `finished` as a special field... not editable on the main record. But I could be wrong. What do you all think about-

- autoset `finished=true` if the satisfied / fulfilled quantity matches or exceeds the requested quantity?
[...]

Issue #2, opened 2019-02-10
Fix JS tests


pospi:
The tests in `holo-rea-dht/tests/` that were generated from the `hc` scaffolding command do not work- it seems they are out of date. Though it is probably much better to do the bulk of our testing in Rust, we would still need integration tests done to validate that the Holochain Zome and Entries are wired up to the core properly.

The up-to-date, working tests can be found [in the `app_spec` folder](https://github.com/holochain/holochain-rust/tree/develop/app_spec) in the HC Rust repo.
Note that we may also have to update the linked Holochain deps in our `Cargo.toml` files, as they are currently set to a specific revision to ensure we have a predictable codebase.

pospi:
Working test setup now in https://github.com/pospi/holo-threaded-comments courtesy of @willemolding

pospi:
In addition, `test/__init.js` needs to be updated to generate multiple container configurations for the multiple app DHTs, and multiple agents which will be needed during testing. Ideally, only the resources necessary to test a given scenario are instantiated (i.e. single-agent tests will instantiate only one agent's containers).

pospi:
Just updating this issue title to refect the current reality.

pospi:
All working, though experiencing some issues with Diorama still.

Failing tests are as follows:

- [ ] updates with fields ommitted leave original value intact
- [ ] links can be written and read between DNAs
- [ ] links can be written and read between zomes

We also need to add tests for:

- [ ] creating, deleting and re-creating a record results in a new record with a different hash (requires implementing creation timestamps)
- [ ] deleting a record of a different type via a deletion method is not possible

And pass these currently disabled tests, when Diorama supports the assertions we need:

- [ ] cross-zome linking errors if target is not present

pospi:
There are now a series of unstarted tests in `economic-resource/resource_logic.js` that need to be completed once the Specification DNA has been implemented.

pospi:
FWIW there are now a series of unstarted tests in `economic-resource/resource_logic.js` that need to be completed once the Specification DNA has been implemented.

pospi:
^`resource_logic.js` tests are fine now, with the exception of #97.

PR #27, opened 2019-06-13
Feature/2 multi dna integration tests

pospi:
Given the breaking API changes in this update, it's probably best to merge this and deal with the churn before we write more code. `link_entries` has changed and requires a link type parameter now, so other updates will be needed after merging into your #19 feature branch.

Also note that the tests are wired up & working, but don't assert anything yet as the backend hasn't had enough implemented to test much.

Testing & integrating this branch will require updating your Holochain framework to `v0.0.19-alpha1`. See the instructions at the bottom of `CONTRIBUTORS.md` on how to do this (steps #2 & 3). Not sure if you're using NIX, but if so it would be great to have separate instructions on how to update via that method.


pospi:
Something else to note about this update- `master` will compile again after merging as I rolled back some of my newtype experiments to avoid craziness until the HDK gets more usable, as per https://github.com/holochain/holochain-rust/issues/1499

pospi:
lol lucky my use of '# 2' above coincidentally matches the issue ID related to this task. I think in future the branch names need to have the hash in them as well for GH to pick up on the relationship automatically.

Issue #23, opened 2019-06-03
Create polished abstractions for graph-like record storage


pospi:
This basically means bringing the work @sqykly has been doing on reproducing `LinkRepo` into the shell of the end-to-end system and validating the architecture by implementing our first inter-DNA bidirectional link.

Most of the groundwork and underlying architecture will be similar to [the structure shown in these diagrams](https://miro.com/app/board/o9J_kx0H2NA=/?moveToWidget=3074457346632648725), and most questions needed per #3 have been answered. Additionally, it is also clear that if one defines a "record" as a single logical entity within a Holochain app then there are actually multiple DNA primitives involved in [constructing that representation](https://miro.com/app/board/o9J_kx0H2NA=/?moveToWidget=3074457346633251435).

It has been observed and discussed that there are some common emerging patterns for record management on Holochain:

- @sqykly referred to these as *"link constraints"* in discussion of whether links should have a *"bidirectional"* attribute added as a *"constraint"* when building the link index.
- @pospi calls these *"link components"* or *"link behaviours"* and wants to use them as low-level declarative building blocks for building up record storage & query structures. We can work towards these simply by wrapping up common functionality as we go. @pdaoust has also expressed a desire to build up 'small composable helper functions' out of what the HDK provides, we hope eventually this process will yield a new set of simplified methods in `hdk::utils`.
- @lucksus has spoken about the extra utility to be found in aggressive decomposition of Holochain app modules into smaller chunks- for example, a separate *'link broadcasting'* DNA which tracks the linkages between records in two cooperating DNAs would allow inter-network connectivity to be analysed and broadcast the appearance of new networks into the ecosystem.
- We've also discussed Holodex with @artbrock, which was a standalone DNA for indexing content on *other* DNAs (kinda like plugging in ElasticSearch in a traditional cluster); and it seems clear that some features will need to interact with bridged DNAs whilst others will be internal logic.

These things being the case, the evolutionary pattern is likely to be:

1. higher-order wrapper functions emerge to take care of managing common data structures (in our case, graph databases), and are integrated into DNA zome handler code
2. some low-level utility DNAs will also emerge, which require external state to be managed in independent networks by way of bridging; in addition to integrating with other DNA code
    - these 'utility' DNAs are likely to have separate user interfaces (in the case of Holodex, a search UI; in the case of link broadcasting, a network map)
    - also likely to have pre-deployed 'recommended' public networks (eg. "the hApp of hApps")
[...]

Issue #22, opened 2019-06-03
Create macros to minimise boilerplate when defining new records & zomes


pospi:
Once #18, #19, #20 & #21 are completed we will have covered most of the data structures needed for VF records. This would be a good time to take the thus-far unavoidable field duplication and verbose `impl` definitions in the core VF modules and turn those into macros so that the code is kept as DRY as possible.

sqykly:
Rather than working this into the spec branch, this probably wants its own branch.  Agreed?

pospi:
Yes of course. Ideally every issue should have its own branch.

pospi:
For clarity: this is about reducing the boilerplate *within each VF record definition file* by taking out the repetition currently needed in defining entry / create request / update request / response structs and the logic inbetween them. It does not cover pulling common field definitions out to reduce the boilerplate needed in eg. defining `note` for each separate VF record.

A good way to approach this is probably to come to agreement on what the final record definition should look like, and then work backwards from there to implement the macro which will translate that into something similar to the current verbose record definition code.

pospi:
Something else to note: we are likely moving to a pattern where referenced fields have separate methods for manipulating them, rather than being editable as fields of the main record structures.

As such, it's likely that the only wiring of link fields into a record is in the READ phase. It's also likely that we will want other types of non-editable "derived" fields in addition to link-based ones (eg `deleteable`).

All this being the case, my suggestion would be to implement derived fields as callbacks which accept the input `Address` and return whatever field data is needed. This would basically replace all zome-layer calls to `construct_response` with a call to some new record-layer method that knows how to read the links internally. But let's be sure we are happy with the revised update logic before doing that since it impacts the ability to construct a response from the input links in a CREATE request...

pospi:
Looking good after implementing `hdk_records` and `hdk_semantic_indexes`- pretty much all zome-specific logic is now specific to that datatype and avoids repetition of basic CRUD and indexing logic.

Issue #20, opened 2019-06-03
Implement bidirectional links between Event.fulfills & Commitment.fulfilledBy


pospi:
Following on from #19, the pattern needs to be extended to work for inter-DNA field links. It is difficult to say what the correct flow of data is for managing updates across DNAs, but one goal is for the `Fulfillment` record to exist on both sides of the network boundary. Ideally fulfillments can be managed in either network and changes would flow between them, but it may be more appropriate to only have one side of the relationship editable. We could start there and see what the added complexity in bidirectional editing looks like?

fosterlynn:
>it may be more appropriate to only have one side of the relationship editable

The natural side to be editable is the "bottom" or "later" side.  For intents/commitments, that is commitment.  For commitments/events, that is event.  My gut is that it should be editable from only the one side.

sqykly:
`LinkRules` is coming and will be a perfect place to refine this idea so that we can re-use it effortlessly as a pattern any time the need arises.

Why is it a goal to have the Fulfillment on both sides?  What makes it different from any other record?  Doesn't that present a security flaw, as you couldn't validate the fulfillment on the same DHT that has all the other records?

pospi:
Not sure what you are meaning by the security flaw... you'll have to unpack that a bit more. At the end of the day if you have two different independent networks with dependent data split between them, either side can't *really* trust the other no matter which way you cut it. But what one network *can* do is facilitate being linked to by data in other networks. That means it basically has to trust / not care about the validity of the IDs of the things linking "in"; or rather, *presume* trust in that the cooperating service linking to one's own is performing honestly. In the majority of our use-cases it might even be that all such linkages are defined at installation time.

Or, think of it this way- you're linking together a bunch of apps that *you* control with identities that *you* own, and you trust those apps / networks to behave as expected in the same way that you trust the GNU command line tools to behave as expected (ie. with code signing and web of trust).

[...]

Issue #18, opened 2019-06-03
Implement 'stable ID' record pattern


pospi:
We first need to implement a wrapper for records that allows other DNAs to detect multiple versions of the same entry as one. This allows cross-DNA linking to work predictably, but requires some wrappers to manage the underlying DHT structures in order to do so.

Basically, this means:

- de-referencing an entry's address the first time it is created and linking the actual entry data off there via an `initial_entry` link.
- creating a read wrapper function which will load the most recent version of some such record structure from the DHT by following the previously created link.
- creating an update wrapper function which follows the `initial_entry` link and updates the de-referenced entry instead of the ID entry
- creating a delete wrapper function which knows to delete both the de-referenced entry ID and linked data

Tests should prove that creating & updating a record does not result in loss of data linked to the entry ID.

sqykly:
Whoops, I was just about to write the initial entry as `original_hash`.  `initial_entry` is fine, though.

pospi:
Not really fussed about internal names. Call things what feels good (:

pospi:
Steps to implementation:

- create non-DRY implementation for each structure type for a VF object: "record", "create payload", "update payload", "entry" & link definitions.
- pull up commonalities into a macro to make defining record *fields* DRY
- (may defer until later) add ability to make defining record fields *between types* DRY

PR #28, opened 2019-06-20
#18 stable id pattern

pospi:
Completes #18 and some other low-level behaviours. This PR:

- implements dereferencing for stable record IDs (referred to as "HoloPointer" by @sqykly), but is done directly in the zome callback handlers for now. Will leave proper encapsulation to @sqykly in the `LinkRules` work of #19 or possibly defer until everything is cleaned up in #23.
- implements expected update behaviour for records (setting fields to `null` clears them, not providing them leaves them untouched)
- removes `Option` wrappers from our type aliases so that `MaybeUndefined` types can be used at the API gateway
- adds some nice accessors for request struct data so that borrow errors & unnecessary field cloning is avoided
- finalises response serialization logic so that records are returned in camelCase and empty fields are ommitted
- adds JavaScript integration tests for low-level core behaviours based on the `EconomicEvent` record (be aware of https://github.com/holochain/diorama/issues/11 when implementing the same)

It does not:

- deal with any links between records, that's for `LinkRules` to take care of in #19
- de-duplicate struct definitions for records- `economic_event.rs` is deliberately non-DRY so that the Commitment structs can be implemented with similar verbosity in #19 & cross-DNA links can be implemented in #20 before we wrap up the common abstractions in #22.


pospi:
Main point of feedback I'm looking for here is whether we think the provided tests are sufficient- if I've missed anything, let me know and I will implement (:

pospi:
Check https://github.com/holochain/diorama for the API.. it's a work in progress, but not well documented yet. S is the holochain adapter, T is the `tape` test harness instance.

PR #33, opened 2019-07-09
update postinstall script for Nix installation if Nix is found

pospi:


pospi:
This has been updated to get a running system as of Holochain `v0.0.27-alpha1`.

Some tests are failing, but the reasons as to why are yet unknown- see https://github.com/holochain/holochain-rust/issues/1662. There is a strong possibility these will be fixed in subsequent Holochain updates; either way the system is in a working state and we shouldn't let these delay a necessary upgrade.

pospi:
To note for @sqykly - you should be able to resolve any Rust crate version mismatch issues after upgrading now, using the officially supported Nix command linked to `npm run clean:modules`.

pospi:
I should have explicitly pointed out- this is ready to merge if you're happy with the changes and are able to get a working system spun up using the instructions in the README.

pospi:
Merging at @fosterlynn's suggestion, will probably continue to self-merge from here as waiting for others who are unavailable is blocking progress.

Issue #58, opened 2019-09-02
Sign off on contributor guidelines


pospi:
I just wanted to draw this out into an explicit decision- I've added a section on [contribution agreements](https://github.com/holo-rea/holo-rea/blob/8198238/CONTRIBUTORS.md#contributor-workflow--coordination-protocol) to the contributors doc in the repo in response to more interested third-parties and potential contributors reaching out.

Since you two are the other major stakeholders in this at the moment I wanted to run it by your eyes and ask for any input; at least one other person should OK this before we start living by it. Feel free to provide suggestions & comments here, or as PRs.

fosterlynn:
Looks OK to me.

pospi:
Closing this- @JhonatanHern is up and running

Issue #57, opened 2019-09-02
Investigate using signalling API to wire up cross-DNA data mirroring


pospi:
At present, data that lives "between" networks (eg. `Satisfaction` & `Fulfillment`) requires a "context record" to be kept on either side of the network boundary in order for participants from both networks to have equal access to information.

You can see this implemented in our system where the [satisfaction zome in the planning DNA pings](https://github.com/holo-rea/holo-rea/blob/c331073/happs/planning/zomes/satisfaction/code/src/satisfaction_requests.rs#L125) the satisfaction zome [in the observation DNA](https://github.com/holo-rea/holo-rea/blob/b1f0886/happs/observation/zomes/satisfaction/code/src/satisfaction_requests.rs#L79) in order to create a duplicate of the record.

We currently only allow the creation of satisfactions & fufillments within the planning DNA space, though ideally both actions should be triggerable via the observation DNA as well. In the current implementation we can't do this, because [you can't configure cyclic bridge dependencies](https://forum.holochain.org/t/bidirectional-bridging/705). But it does seem like a common practise to have logic in one DNA responding to activity in another.

The signalling API may be the answer to this, provided signals emitted in one DNA are received by all connected DNAs running in the conductor; and no bridging is needed to enable this link.

If we can get this working we should remove the unnecessary bridge and provide for an ability to create fulfillments in the observation DNA as well. We should also decide on some standard message structures in order to wire up similar triggers between inter-network parts of the system.

pdaoust:
@pospi is the dependency graph something you could scribble up a quick diagram of? Just so's I can get a grasp of what should be responding to what.

pospi:
Consider two DNAs: one containing posts and another containing comments. In this publishing platform, posts may link to comments and comments may link to posts. But multiple separate overlapping comment DNAs may interact with the same posts DNA. On both side of the links, the DNA users want to retain a record of where the links were made to, when they were made, what access they were provided under, etc- call this payload "the [XDI metadata](https://en.wikipedia.org/wiki/XDI)".

Does that help to illustrate the general case at all without requiring an actual illustration? Not near a decent screen atm

pdaoust:
[...]

Issue #42, opened 2019-08-16
Finalise fulfillment & satisfaction data structures


pospi:
These need to provide for the intermediary records (they currently link directly between `EconomicEvent`, `Commitment` and `Intent`) and all CRUD methods as deemed appropriate by https://github.com/holo-rea/holo-rea/issues/37.

Once finalised we can roll this pattern out to all zome API handlers and complete the Observation and Planning modules.

fosterlynn:
Made some changes to those records in the current VF release, fyi.  Should be stable for now from that side.

pospi:
@fosterlynn I'm going to have to set this up so that fulfillments can only be managed via the planning DNA for now... does that sound like the most sensible course of action?

Can't have it work both ways- see https://forum.holochain.org/t/bidirectional-bridging/705

pospi:
I've actually done it the other way around, which I think makes more sense. A participant has no reason to log fulfillments if not for commitments they are aware of, correct? So should it not be something that is done in the planning DNA?

I think I see it as "fulfilling a commitment" is an action which happens in the `planning` module, and `observation` modules are notified (retroactively) by planning modules that their events have gone to fulfill other commitments. 

To be clear- the records end up on both sides. [See complete test case](https://github.com/holo-rea/holo-rea/blob/d1cb52a/test/core-architecture/test_links_cross_dna.js#L13) to get a sense of the data and links involved.

pospi:
I think it should be bidirectional, ideally- participants with access to either network should be able to trigger saving the fulfillment. But we'll see what the core team says about cyclic dependencies in DNA bridging configurations.

fosterlynn:
>I think it should be bidirectional, ideally- participants with access to either network should be able to trigger saving the fulfillment.

What is the use case you imagine for coming at it from the commitment side instead of when logging an event?  And what is the use case when would someone saving a fulfillment not have access to both observation and planning?

pospi:
I think I'll split this out and log as a separate issue when it becomes an important detail. For now it's a minor technical thing- someone saving a fulfillment should always have access to both DNAs; the concern is only for others reading.

The basic implementation is more or less complete now, I just need to loosen the validation on the `Satisfaction satisfiedBy Commitment` links, since they are optional (can also reference an `EconomicEvent` in a foreign DNA). Expect a PR tomorrow some time.

fosterlynn:
[...]

Issue #39, opened 2019-08-16
Finalise optimal type guarantees & code quality for Multi-DNA PoC


pospi:
- [ ] wire up newtypes to records (`vf_core/src/type_aliases.rs`)
- [ ] wire up inbuilt actions to records (`vf_knowledge/src/action/builtins.rs` & record CRUD structs). For future plans, see #38.
- [ ] move `default_false` serde field filler out of record classes into `hdk_graph_helpers`
- [ ] return errors instead of discarding with `get_links_and_load_type`
- [ ] check that all remaining record fields for 3 main flow classes (event, intent & commitment) are done

pospi:
> wire up newtypes to records & through zome API gateway

Note for when running through this: the latest implemented zome is likely to be the best one to follow as an idiomatic pattern. Some structures (particularly use of cloned data vs ref passing) have evolved over time, and there isn't always time to comb back over previous zome code.

At time of writing this is `Process`, but others may have superceded it by the time this technical debt is paid off.

Connoropolous:
I added the double check me label since this is from 2019 and a lot of the code references look outdated

PR #59, opened 2019-09-02
Fulfillment & satisfaction APIs

pospi:
This is mainly focused on completing #42 and passing end-to-end integration tests for both `Fulfillment` and `Satisfaction`-related functionality. It includes everything needed for a working first-pass at all flow records (these plus `EconomicEvent`, `Commitment` and `Intent`) and has full support in GraphQL for the related "read by ID" and relationship traversal queries.

It also adds the following updates:

- Adds [get_links_and_load_entry_data](https://github.com/holo-rea/holo-rea/blob/39f699c/lib/hdk_graph_helpers/src/link_helpers.rs#L202) helper, for reading a set of linked entries and base addresses attached to some context entry.
- Adds [read_from_zome](https://github.com/holo-rea/holo-rea/blob/c0173cf/lib/hdk_graph_helpers/src/record_helpers.rs#L154) helper to wrap up error handling & response decoding for inter-DNA RPC calls
- More integration tests- list in #2 is up to date
- Bugfixes to `MaybeUndefined` to allow for serialisation & request passing
- Better error messages from API deletion handlers
- Cleanup of zome API handler parameters and optimisation of memory usage
- Formalised contributor agreements a bit more (see #58)

**To do:**

There are a few additional cleanup tasks added to #39 - nothing major, just cleanup. My goal is to get the base system passing integration tests ASAP so that others can start developing on it before we make a serious optimisation pass. We also need to think through what we have so far and sound out any worthwhile improvements before continuing on to other modules besides observation, planning & specification.

This work also opens up some other avenues for exploration around creating shared abstractions for storing link metadata. It is quite likely that adopting an [XDI](https://en.wikipedia.org/wiki/XDI) format as a base for records like `Fulfillment` and `Satisfaction` would be beneficial for interoperability with projects like the [Link Index DNA](https://github.com/holo-rea/ecosystem/wiki/Modules-in-the-HoloREA-framework#links).

It may also be worth investigating the Holochain signalling API more as a followup- see #57.


pospi:
I know you're all on the road right now so I'm going to self-merge this so that @philipbeadle & other devs can get started on an up-to-date system this/next week (:

Issue #54, opened 2019-08-26
Implement Process


pospi:
Nothing crazy here, just another record type with CRUD routes to define in the `observation` DNA.

PR #62, opened 2019-09-11
Feature/54 process api

pospi:
Implements #54. Also:

- finished wiring up newtype field aliases through the API (see #39). Still needs a bit of a code review, lots of data cloning going on in there, and awkward conversions from the aliased form to `Address`. But the external API behaves in an efficient way, which is what matters.
- new patterns emerging for different index handling capabilities have been implemented in new link management methods in `hdk_graph_helpers`, this provides for local & remote `input_of` etc links between processes & flow records
- split RPC related methods out into new `hdk_graph_helpers` sub-module & standardised a lowest-common-denominator index synchronisation protocol for managing links between entries in remote DNAs
- some (not all) updates to the core record types to align with latest `vf-graphql` spec
- more integration tests. Updated #2 and moved the todo list into the first comment so that progress shows in the issue list.

I'm just going to keep self-merging these until @fosterlynn is back from vacation next week, so that things are in a good place for a code review when @philipbeadle gets on board (also next week).


pospi:
Oh, and the other celebration here is that all implemented tests now pass!

Issue #65, opened 2019-09-16
Details of update logic for EconomicResources?


pospi:
I've got some questions on how to manage the fields of an `EconomicResource` in regard to event-related functionality through the API, apologies if they are documented elsewhere; if not- this might be a good opportunity to describe the state machine at a high level.

When **creating** a resource via the additional `createResource` parameters in the `createEconomicEvent` GraphQL mutation:

- `stage`: should this be provided as an input parameter, just start off as *'none'*, or should it take on the `ProcessSpecification` of the related `EconomicEvent` that resulted in the resource creation, if `EconomicEvent.outputOf` is set? Or `inputOf`? Or something else?
- `unitOfEffort`: similarly, must this be explicitly provided or does it come from the associated `ResourceSpecification`?
- and `state` comes from the event's `action`, if the action is `pass` or `fail`? If that's going in to the model... how does a "pass" expire for inspections that need periodical review? It doesn't feel right for it to "stick" to the resource...
- I presume `accountingQuantity` adjusts in response to `e.resourceQuantity` based on `e.action.resourceEffect`? (being either `increment` or `decrement`)
- are any other `EconomicResource` fields affected by or derived from fields of the associated `EconomicEvent` that is provided at creation time?
- Should the resource be linked to the event responsible for creating it somehow? I suppose we would transparently create the resource address as part of the transaction and inject that as `resourceInventoriedAs` in the stored event?

I also have outstanding ambivalence about `createResource` as a parameter name in the event API, even though we've discussed that it's unavoidable. What if we renamed it to `observedResource`? I'd feel better about that :P

For **updating** a resource directly (pretty much only provided for correcting data entry errors...)

- Are we missing a bunch of fields in the mutations API to mirror the new resource fields in 0.3, or are some omissions intentional? (`lot`, `unitOfEffort`,  `stage`,  `state` &  `containedIn`)

For **updating** a resource in response to an event:

- Is `classifiedAs` ever altered by events, or does it retain the value it began with? i.e. do I update `r.classifiedAs` in response to `e.resourceClassifiedAs` if `e.resourceInventoriedAs` is provided? If so, is `e.toResourceInventoriedAs` modified as well?
- Is `onhandQuantity` affected by `e.resourceQuantity`, or just `accountingQuantity`?
- Does `containedIn` respond to any action types?
- What about `currentLocation`?
- Kinda repeats, but- what is the logic for how to update `stage` & `state`?

[...]

Issue #63, opened 2019-09-11
Figure out why certain tests are failing in CI


pospi:
See https://circleci.com/gh/holo-rea/holo-rea/3

They both appear to concern the updating of field indexes between records referenced in different zomes within the same DNA.

Works on my local setup. @fosterlynn @bhaugen when you get set up next week, would be good to see if you can reproduce the problem locally.

PR #66, opened 2019-09-26
Feature/53 economic resource

pospi:
This finishes the bulk of the event <> resource logic detailed in #65 with the exception of initialising resource quantities (pending final decision), and items that depend on #55.

There is still some pending work around deleting resources and handling event updates which I will log as separate issues.

Am going to merge this straight away to keep things moving with integration into our first PoC.


pospi:
The only [tests failing](https://circleci.com/gh/holo-rea/holo-rea/17) are the weird ones from #63. At some point I might try switching to TravisCI and see if that clears things up.

PR #72, opened 2019-10-16
Feature/platform update

pospi:


pospi:
@JhonatanHern you will have to exit your `nix-shell` and re-enter it after pulling these changes. You will also have to update the Rust crate version IDs in any `Cargo.toml` files you've created for the new *specification* DNA.
Also note that you no longer need to update hashes in `conductor-config.yaml` after a code change, so that's less painful (:

PR #77, opened 2019-10-31
add contributors list first pass

pospi:
I've never started one of these before so thought I should run it by the group and give people an opportunity to add websites or any other details they want before merging?

I'm not sold on the headings either, just pulled them from the NRP codebase and added some sections for contributions that didn't already fit.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -1,16 +1,13 @@
-Code contributors
-=================
+Technical contributors
+======================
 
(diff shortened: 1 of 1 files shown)


pospi:
Added various others who have contributed to architectural R&D, as well as more recent consultants & funders. I think we've discussed everything else that's needed here for now... @pdaoust @philipbeadle @willemolding if you'd like to update profile links feel free to send a PR or comment here (;

Issue #68, opened 2019-09-26
Implement event update & delete logic for resources


pospi:
Following on from #53.

Updating resources from newly created events is pretty simple. Altering or removing existing events is more complex, as it requires that the effects of the previously existing event are rolled back before applying changes.

There are a swathe of test cases at the end of `resource_logic.js` which need to be implemented and proven.

pospi:
More info on this, see https://github.com/holo-rea/holo-rea/issues/65#issuecomment-543142930

pospi:
I think we're more or less good on the `EconomicEvent` aspects of this following discussion in #65.

Events can no longer be updated in any way that impacts resources, so this avoids many unnecessary complexities. The logic for how new events affect resources has also been finalized except for the fields which depend on the *specification* module.

This leaves the resource creation itself, and update of its fields (separately to events); see https://github.com/holo-rea/holo-rea/issues/65#issuecomment-549252043

pospi:
Done. Hopefully this is the last one I'll have to self-merge as @JhonatanHern should be back online and able to review merges again (:

PR #81, opened 2019-11-07
Feature/68 resource update logic

pospi:
Implements #68. Note that #76 must be merged first since this work depends on that branch.


pospi:
Found there are some failing tests, some unknown internal error. So this is not quite ready yet...

Issue #98, opened 2019-12-05
Implement Proposal DNA


pospi:
It seems like this might want to be part of a first release, given recent discussions with @fosterlynn about the mechanics of de-listing `Intents` from marketplaces.

The proposed means of de-listing is to set the `endDate` of the related `Proposal` to the current time.

pospi:
@JhonatanHern have been looking through last week's work- 6022727 is totally sensible, and exactly what I would have probably done. I expect you may have copied it from the `ResourceSpecification` indexes in the `EconomicResource` zome, or the `Process` indexes in `Commitment`? Those are actually things I need to clean up- apologies! In building the latter of these I realised that it isn't good practise to put base entry defs for remote entry types in the same zome as the entries they connect to... because, like `Process`, they might connect to more than 1 entry and now I have the `Commitment` zome being dependent on the `Intent` zome in a way that it really shouldn't be.

These links also differ from `ProposedIntent` in that they don't exactly link to each other... it's more like `ProposedIntent` *is* the link between `Proposal` and `Intent`. There is additional metadata in the link, in this case `reciprocal`. Basically in the [graph helpers lib terminology](https://github.com/holo-rea/holo-rea/tree/master/lib/hdk_graph_helpers#theory), the kind of link you have defined here is a "direct remote index" when you actually need an "indirect remote index".

So instead of the `proposed_intent_base_entry_def` in the `Intent` zome, what you need is an additional *planning* zome for `ProposedIntent` (use the `new-zome` Hygen generator to start this). You can use `Fulfillment` or `Satisfaction` as a pattern to follow- instead of one `lib` folder, these zomes have an additional `lib_origin` and `lib_destination`. You will see that the `_origin` side (in this case, the `ProposedIntent` zome in the *proposal* DNA) calls through to the `_destination` side via a DNA bridge. The two library crates actually manage different parts of the index, which you can easily tell by looking at the `_storage_consts` crates they are dependent upon. To get started you'll probably want to move the existing `lib` folder to `lib_origin`, then move the `construct_response()` function back into `lib`. `lib_destination` will contain the code you need to write to manage the `ProposedIntent` records within the *planning* DNA.

[...]

PR #144, opened 2020-03-04
Feature/98 proposal

JhonatanHern:
Branch for the proposal happ and things related to it like external links, tests and resolvers.


happs/proposal/app.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "HoloREA: proposal",
+  "description": "Proposal DNA for the REA economic network",
```
pospi: We need a better app store description from the core VF docs somewhere. @fosterlynn any suggestions?

lib/rea_intent/storage_consts/src/lib.rs
```diff
@@ -15,4 +15,7 @@ pub const INTENT_INPUT_OF_LINK_TAG: &str = "input_of";
 pub const INTENT_OUTPUT_OF_LINK_TYPE: &str = "vf_intent_output_of";
 pub const INTENT_OUTPUT_OF_LINK_TAG: &str = "output_of";
 
+pub const INTENT_PROPOSED_INTENT_PUBLISHES_LINK_TYPE: &str = "vf_intent_proposed_intent_publishes";
+pub const INTENT_PROPOSED_INTENT_PUBLISHES_LINK_TAG: &str = "intent_proposed_intent_publishes";
```
pospi: Change to simply "publishes"

test/proposal/test_proposedintent_crud.js
```diff
@@ -0,0 +1,172 @@
...
+  `, {
+    id: proposalAdress,
+  })
+  t.equal(getResp.data.res.id, proposalAdress, 'proposal fetch succesful')
+  t.equal(getResp.data.res.publishes[0].id, proposedIntentAdress, 'proposedIntent fetching from proposal succesful')
+  t.equal(getResp.data.res.publishes[0].publishes.id, intentAdress, 'intent fetching from proposedIntent succesful')
+
+  await alice.graphQL(`
+    mutation($in: String!) {
+      res: deleteIntent(id: $in)
```
pospi: This is what's making this test cause errors. Now that the `s.consistency()` calls are fixed, the error being reported is what is expected- it is failing when fetching `proposal.publishes.publishes` because you have deleted the `Intent`, but the system doesn't yet account for that and delete the `ProposedIntent`. Will open an issue about deletions separately, but for now, this needs to be changed to triggering / testing `deleteProposedIntent`.

happs/proposal/app.json
fosterlynn: Hmmm. How about something like

"Publication of intents (requests or offers), often with what is expected in return."

JhonatanHern: I'll add it

lib/rea_intent/storage_consts/src/lib.rs
JhonatanHern: done

modules/vf-graphql-holochain/mutations/proposedIntent.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export const proposeIntent: createHandler = async (root, args) => {
+  const adaptedArguments = {
+    proposed_intent: {
+      publishedIn: args.publishedIn,
+      publishes: args.publishes,
+      reciprocal: args.reciprocal
+    }
+  }
+  return createHandler(adaptedArguments)
```
pospi: why not just pass `{ proposed_intent: args }`?

modules/vf-graphql-holochain/mutations/proposedTo.ts
```diff
@@ -0,0 +1,35 @@
...
+type createHandler = (root: any, args) => Promise<ProposedToResponse>
+
+export const proposeTo: createHandler = async (root, args) => {
+  const adaptedArguments = {
+    proposed_to: {
+      proposed: args.proposed,
+      proposedTo: args.proposedTo
+    }
+  }
+  return createHandler(adaptedArguments)
```
pospi: `{ proposed_to: args }`?

modules/vf-graphql-holochain/queries/proposal.ts
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * @package: HoloREA
+ * @since:   2019-09-12
+ */
+
+import { zomeFunction } from '../connection'
+
+import {
+  Proposal,
+  ProposedTo,
```
pospi: Do these types need to be imported, aside from `Proposal`?

test/proposal/test_proposedintent_crud.js
```diff
@@ -0,0 +1,152 @@
...
+        }
+      }
+    }
+  `, {
+    id: proposalAdress,
+  })
+  t.equal(getResp.data.res.id, proposalAdress, 'proposal fetch succesful')
+  t.equal(getResp.data.res.publishes[0].id, proposedIntentAdress, 'proposedIntent fetching from proposal succesful')
+  t.equal(getResp.data.res.publishes[0].publishes.id, intentAdress, 'intent fetching from proposedIntent succesful')
+
```
pospi: There is a test missing here that is needed to finish the `Intent` side of the query. Before deleting the `ProposedIntent`, you also need to ensure that querying the `Intent` by ID and then following it through `Intent.publishedIn.publishedIn` to load the `ProposedIntent` and `Proposal` records works.

As suggested above, it would be good to also create a second `Intent` and `ProposedIntent` so that you can verify that returning multiple records for `Intent.publishedIn` works.

lib/rea_proposed_intent/lib/src/lib.rs
```diff
@@ -0,0 +1,162 @@
+use hdk::{
```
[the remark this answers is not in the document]
JhonatanHern: done, I know that only my files had that error, but only to be sure, I wrote [this](https://gist.github.com/JhonatanHern/ca37a993194049d8096ef47684d2990f) and thought you could find it interesting. It finds the files that do not start with a string.

lib/rea_proposal/lib/src/lib.rs
```diff
@@ -0,0 +1,180 @@
...
+        &new_entry,
+        get_link_fields(base_address),
+    ))
+}
+
+fn handle_query_proposals(params: &QueryParams) -> ZomeApiResult<Vec<ResponseData>> {
+    let mut entries_result: ZomeApiResult<Vec<(ProposalAddress, Option<Entry>)>> =
+        Err(ZomeApiError::Internal("No results found".to_string()));
+
+    match &params.publishes {
```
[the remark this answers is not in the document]
JhonatanHern: I'm going to call `unimplemented!();` for that

lib/rea_proposed_intent/lib/src/lib.rs
```diff
@@ -0,0 +1,162 @@
...
+        "index_publishes",
+        Address::from(PUBLIC_TOKEN.to_string()),
+        INTENT_BASE_ENTRY_TYPE,
+        PROPOSED_INTENT_PUBLISHES_LINK_TYPE,
+        PROPOSED_INTENT_PUBLISHES_LINK_TAG,
+        INTENT_PROPOSED_INTENT_PUBLISHES_LINK_TYPE,
+        INTENT_PROPOSED_INTENT_PUBLISHES_LINK_TAG,
+        base_address.as_ref(),
+        vec![(proposed_intent.publishes.as_ref()).clone()],
+    );
```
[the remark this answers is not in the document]
JhonatanHern: Finally done with this part, just pushed.

happs/planning/zomes/intent/code/src/lib.rs
```diff
@@ -43,6 +49,33 @@ mod rea_intent_zome {
         base_entry_def()
     }
 
+    #[entry_def]
```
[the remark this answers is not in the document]
JhonatanHern: done

pospi:
Implements #98

Issue #148, opened 2020-04-06
Update to align with ValueFlows 0.8.0 spec


pospi:
- [x] remove `AnyType` and add `__resolveType` resolvers for `AccountingScope` (may as well cover all union types while we're at it)
- [x] change use of all query endpoints to remove the `all*` prefix (probably only affects tests)
- [x] fix all deletion methods taking `String` instead of `ID` in queries (probably only affects tests)
- [ ] swap out the `URI` and `DateTime` resolvers; consider using https://github.com/Urigo/graphql-scalars. Add tests to ensure all necessary date formats are valid.

pospi:
Split the last part of this off into #156.

fosterlynn:
What's this?  Is that what you think hte next VF release is?  Should we attempt to realign?

>ValueFlows 0.8.0

fosterlynn:
Deleted last comment - realized you are aligning holo-rea with vf-graphql.

PR #157, opened 2020-06-06
Feature/148 protocol update

pospi:
#148 makes updates for the latest vf-graphql versions.

This involved some backward-incompatible changes to query names and use of `ID` instead of `String` for deletion methods. All implementations will have to update accordingly.

Also updates GraphQL modules throughout the project and shifts GraphQL to an NPM `peerDependency` so that dependent projects may specify the version of GraphQL to use, rather than being lead by the VF/Holo modules.

### DIFF
--- a/apps/holorea-graphql-explorer/package.json
+++ b/apps/holorea-graphql-explorer/package.json
@@ -13,11 +13,10 @@
     "react-dom": "^16.13.1"
   },
   "devDependencies": {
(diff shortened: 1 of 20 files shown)


pospi:
Would be good to get a peer review from either you guys or @sdelvalle57 as these are the most disruptive changes in a while, but happy to self-merge if there isn't bandwidth for it.

PR #164, opened 2020-07-21
Feature/switch to pnpm

pospi:
I'm moving away from use of Yarn as a package manager, towards https://pnpm.js.org

It has a nicer directory structure which avoids some of the issues that Yarn runs in to when sharing modules between workspaces- in this case, Holo-REA and [apps](https://github.com/holo-rea/?q=app-) which depend on the `@valueflows/vf-graphql-holochain` package; as well as other wiring between this module and `@valueflows/vf-graphql`, and even sharing components directly  between some repos.

To make all this work correctly the basic premise is that you have all of the projects cloned as siblings in the same parent directory. There is a [metarepo](https://github.com/holo-rea/valueflows-project-metarepo) for this, which now has a simpler setup script to [manage the linking](https://github.com/holo-rea/valueflows-project-metarepo/blob/48e3a5b/init.sh#L99) using PNPM's simplified `link` command.

Anyway, `npm run clean:modules` after checking out this version, try setting up the repo again and follow the prompts, if all goes well you should end up with an install managed by PNPM.

PR #165, opened 2020-07-22
Feature/update holochain50

pospi:
**Please merge #164 first!**

Updates to latest Holochain, has some reasonable performance improvements. After checking out, you will need to exit and re-enter Nix shell to update the platform.


oro13 (approved):
`pnpm run test:integration` checks out

Issue #174, opened 2021-09-29
HDK3 cleanup tasks


pospi:
The migration to "Holochain RSM" is now complete-enough to be considered a new stable foundation to integrate from. A prototype version compatible with the previous HDK2 is tagged at the [hdk-2.x branch](https://github.com/holo-rea/holo-rea/tree/hdk-2.x).

Filing this issue to track cleanup jobs needed to finalise the task and achieve feature parity with the old version-

### Blocking

- [x] #213
- [x] #218
- [x] get GraphQL explorer wired up again (pending resolution of above)
    - [x] update conductor running command

### Bugs

- [x] fix all `EconomicResource` container refs being cleared when only 1 is removed
- [x] ensure `EconomicResource` stage & state still work (they query the DHT for related records directly...)
- [x] ensure unit CRUD (and "anchored records") work as expected after changes to indexing architecture

### Code clarity & maintainability / repo organisation / cleanup

- [x] #216
- [x] #267
- [ ] #214
- [ ] #215
- [x] move all indexing library code to its own crates, and split out index helpers into own `hdk_semantic_indexes` crate that `hdk_records` can depend upon
- [x] change all `generate_query_handler` methods to a more abstract macro that can be statically compiled
- [x] simplify / consolidate naming of indexing helper methods, since all index types now have a coherent API
- [x] cleanup some residual zome API code duplication & misplaced structs:
    - WASM API parameter structs should be declared in `rpc/`
    - `ByHeader` & `ByAddress` only declared once; `ByHeader.revision_id` instead of `.address`

### Administration & documentation

- publish Nodejs modules
    - [x] `@valueflows/vf-graphql-holochain`
    - [x] `@vf-ui/graphql-client-holochain`
- [x] #70 
- update workflow automation docs
- update [DNA Auth Resolver](https://github.com/holochain-open-dev/dna-auth-resolver/) readme to point to final location of `hdk_records` lib, publish `hc_zome_dna_auth_resolver` crates

### Work ahead and other things to note

- This upgrade brings us a long way towards #49 as all record identifiers now include a `DnaHash`
- There's stuff to leverage in terms of the Agent module, eg. #172
- Revisit ID field encoding someday #173

artbrock:
## RE: Blocking Issues

### hREA is too large to install (core issue)
 - Unable to install large apps holochain/holochain#966
 
[...]

Issue #43, opened 2019-08-16
Create unit test suite covering core module dependencies


pospi:
(observation, planning & specification)

This refers to Rust unit tests- currently the only module with unit tests is `hdk_graph_helpers::maybe_undefined`.

pospi:
- [ ] finish Rust-level test coverage & mocking for all crates in the `lib/` directory
- [ ] reproduce the assertions in `test/core-architecture` as cargo tests, and remove the JS-level ones

https://github.com/holochain-open-dev/minimal-hdk-unit-test-example and https://forum.holochain.org/t/unit-testing-zomes-with-mockhdk/5421 are probably useful resources.

PR #195, opened 2022-02-22
Run test suite in GH Actions

harlantwood:
- [x] delete old circle and travis stuff

Runs in Ubuntu and Mac.  Would run in Windows too, but I don't think nix works there.

I also came across these lines in circle and travis configs (circle has not run for 2 years, travis I'm not sure) -- maybe these are useful ways to run the tests?  @pospi LMK if you'd like different test commands, or if you want test in GH actions at all.

```
nix-shell --run 'npm i -g pnpm' && nix-shell --run 'npm run dht:sim2h &'
nix-shell --run 'pnpm install --network-concurrency 1 && npm run build && npm run test:integration:test'
```

```
nix-shell --run 'pnpm install && npm run build && npm run test:integration:test'
```


harlantwood:
Update: looks like travis last ran 2 years ago as well as Circle.  Links:

https://app.circleci.com/pipelines/github/holo-rea/holo-rea
https://travis-ci.org/github/holo-rea/holo-rea

pospi:
@Connoropolous actually just submitted a PR for github actions, and those are enabled now to generate builds.

Whether we want the test suite to run as part of that process is another matter, possibly an evolving one. At present still maybe 40-50% of the test suite is failing, mostly just because they haven't yet been updated for RSM (see #174). So that's why I haven't yet re-enabled Travis.

There might be a worthwhile interim step where we run tests on one CI system and builds on another. Tests should be run for any change on `sprout`, but builds only get run if you push a version tag. It's possibly also worthwhile keeping them as parallel things on different infrastructure because it means, well, parallelisation.

Eventually though yes I think it's best if everything gets run in GH actions and the build doesn't get kicked off if the tests fail!

Oh, and a note on the test suite- I didn't have many Rust tests in the beginning so was only running Tryorama tests in JS. Eventually I think it would be best to simply run `npm run test`, which first runs unit tests with `cargo test` before triggering the integration test suite in the `test/` subdirectory. Rust tests probably need a QA and some polish before doing that, and the stuff in `test/core-architecture` should be being done with cargo rather than nodejs & tape (see #43).

Does that help? Opinions & input very welcome (:

Connoropolous:
I still haven't sent Pospi the cachic secret, that we can add as a secret to the repository, so I will do that now.

Connoropolous:
I also just approved the CI runs for this PR, since Harlan and this fork are 'new contributors'

Connoropolous:
myself I've never yet tried running the test suite, so I can't comment like Pospi can

.github/workflows/test.yml
```diff
@@ -0,0 +1,68 @@
...
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            .cargo/bin/
+            .cargo/registry/index/
+            .cargo/registry/cache/
+            .cargo/git/db/
+            target/
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
```
Connoropolous, line 30: this repo has the rather anomalous behavior of not checking the `Cargo.lock` file into git, which as I understand it has its reasons, but from my experience has had some real downsides

Connoropolous: in my other one, I use `Cargo.toml`, while realizing that that's not best practice

harlantwood:
@pospi have you considered moving to sweetest instead of upgrading
tryorama?  It’s pretty great to have your tests of rust lib in rust.  It’s
slow (unless you’re on an M1 mac) but worth it in my experience so far.

harlantwood:
> Tests should be run for any change on sprout, but builds only get run if you push a version tag. 

That’s exactly the behavior of this PR was merged.  And they would run in parallel if you pushed a tag. 

> Eventually though yes I think it's best if everything gets run in GH actions and the build doesn't get kicked off if the tests fail!

Sure, we could chain the builds conditionally later.

harlantwood:
> I still haven't sent Pospi the cachic secret, that we can add as a secret to the repository, so I will do that now.

Ah cool.  Still the run was about 5 min, must have been found in the holochain-ci cache.

.github/workflows/test.yml
harlantwood: > this repo has the rather anomalous behavior of not checking the `Cargo.lock` file into git, which as I understand it has its reasons, but from my experience has had some real downsides

Hmm I think this is actually best practice for libraries (but not binaries) — https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries

harlantwood: > in my other one, I use `Cargo.toml`, while realizing that that's not best practice

Thanks, should update

harlantwood:
@pospi  overall I’d suggest you just leave this open until you’re ready to run tests in CI, happy to help more at that time if desired.

Connoropolous:
"have you considered moving to sweetest instead of upgrading
tryorama?  It’s pretty great to have your tests of rust lib in rust.  It’s
slow (unless you’re on an M1 mac) but worth it in my experience so far."
I am speaking out of turn, but I do know that the tests actually test the graphql part of the stack as well, so it's not useful to switch off of tryorama

.github/workflows/test.yml
Connoropolous: That kinda makes sense. The zomes themselves are more like binaries, while the things in `lib` are libraries, so 🤷 I'm not sure

.github/workflows/test.yml
```diff
@@ -0,0 +1,68 @@
...
+        if: ${{ runner.os == 'macOS' }}
+        run: sw_vers
+
+      - name: Set up nix
+        uses: cachix/install-nix-action@v16
+        with:
+          nix_path: nixpkgs=channel:nixos-21.05
+          extra_nix_config: |
+            substituters = https://cache.nixos.org https://cache.holo.host https://ci-builds.cachix.org https://holochain-ci.cachix.org
+            trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= cache.holo.host-1:lNXIXtJgS9Iuw4Cu6X0HINLu9sTfcjEntnrgwMQIMcE= cache.holo.host-2:ZJCkX3AUYZ8soxTLfTb60g+F3MkWD7hkH9y8CgqwhDQ= ci-builds.cachix.org-1:fxB0+h/MMlCpXf6hFsQM31YpHbaQoRmcNPNHwDUkXA4= holochain-ci.cachix.org-1:5IUSkZc0aoRS53rfkvH9Kid40NpyjwCMCzwRTXy+QN8=
```
Connoropolous, line 50: we should consolidate between this way of defining these, and the `nix.conf` way that is currently at play. I have zero attachment to which way it goes

Connoropolous: I am referring to this file, and the command in the Release CI workflow that references it
https://github.com/holo-rea/holo-rea/blob/sprout/.github/nix.conf

.github/workflows/test.yml
harlantwood: > That kinda makes sense. The zomes themselves are more like binaries, while the things in `lib` are libraries, so 🤷 I'm not sure

More nuanced suggestion here: https://stackoverflow.com/a/63727523

pospi:
> @pospi have you considered moving to sweetest instead of upgrading tryorama? It’s pretty great to have your tests of rust lib in rust. It’s slow (unless you’re on an M1 mac) but worth it in my experience so far.

I have, but the decision not to is based on the layer I'm intentionally testing at. There are still tests that use the raw `$cell.call()` methods, but I consider that pattern deprecated and in need of a refactor- ideally all tests should be running through the GraphQL API, because this means the testing is being done against the domain model rather than the Holochain API interface; and in the way that GraphQL binds schemas to the backend you get automatic checking that your "frontend" & "backend" schemas are aligned.

There are some other nice benefits to this in that the test become "how-tos" for prospective developers (the GraphQL is readily discernable among the JavaScript test boilerplate); and the tests can be used against other systems implementing the [ValueFlows GraphQL spec](https://lab.allmende.io/valueflows/vf-schemas/vf-graphql).

pospi:
There are also a few extra bits of serialization logic and module binding to other Holochain components in the GraphQL layer and so for hREA UI apps it's best to consider that as the outer boundary of the system. If you're building Rust apps that interact with Holochain directly, then you're looking at more direct interaction with those components anyway and understanding / binding directly to those APIs. In that case, the code in the GraphQL layer for modules [such as agent](https://github.com/holo-rea/holo-rea/blob/sprout/modules/vf-graphql-holochain/queries/agent.ts) might be useful in replicating the logic for adapting standard Holochain concepts & modules to the REA vocabulary.

So far, [agent registration](https://github.com/holochain-open-dev/agent-registration/) and [DNA auth resolver](https://github.com/holochain-open-dev/dna-auth-resolver/) are the only two zome libraries hREA is directly dependent upon, and the way they're included & interact with Rust's crate system is worth exploring in the context of above as I think there could be answers to some rough edges.

> That kinda makes sense. The zomes themselves are more like binaries, while the things in `lib` are libraries, so shrug I'm not sure

Nor am I. The canonical way for reusing code in hApps is something I haven't revised in a while, but there might be a solution if integrated with GH release builds.

- **question:** can we publish zome WASMs to the release pages? (don't see why not...)
- **question:** can we use `url` to configure DNA bundles in place of `bundled`?

This would solve a few issues, expressed by the non-rea zome libraries in `zomes/`. The best practise for including other zome code in an app at the time was to re-export the original crate, [like this](https://github.com/holo-rea/holo-rea/blob/sprout/zomes/dna_auth_resolver/src/lib.rs). The issue becomes unnecessary [duplication / derivation of the original crate name](https://github.com/holo-rea/holo-rea/blob/sprout/zomes/dna_auth_resolver/Cargo.toml#L2), resulting in a different zome hash, losing the ability to provably show that zomes being imported into hREA are unmodified standard modules. My thinking is we could use `url` refs to point to pre-baked WASMs and obviate the need to create these intermediary crates (and make room for updating that process to a Holochain-native deployment method via [DevHub](https://github.com/holochain/devhub-dnas/) at some point).

pospi:
The other unanswered question I have on the topic of authoring crates to behave like libraries and binaries is how to do this in a way that Cargo doesn't end up getting confused about.

Currently, the crates compromising the core hREA framework are pinned loosely (`hdk = "0.0"`); and only the zome crates (binaries with `crate_type = "cdylib"`) are pinned to specific HDK versions. The thinking was that this would allow others to import the framework without having to manage niggling dependencies and ask upstream crate authors to bump their HDK dependencies in time with yours, which was what was happening previously and somewhat drained the energy from complex projects in administrative time. So if we can solve some of this above stuff for zome library linking we can remove the last remaining parts of this version coupling issue and DNA authors will be able to mix WASMs compiled for different versions of the HDK without issue so long as those crates are compiled to compatible versions of HDK (which they should be, once HDK moves out of alpha and into proper semver).

Something I've not yet resolved with this is how to correctly set my dependencies across crates to *guarantee* that Cargo will not check out duplicate versions of the HDK and start to create build errors out of duplicate / mismatched versions of the same struct definitions.

Adding `Cargo.lock` to this repo seemed to be a good idea, as it prevented the Github release scripts from getting confused about the correct versions of `hdk` to use with building different parts of the codebase. I *think* what is happening now is that builds run from the hREA project dir use its lockfile as priority and this triggers Cargo to use the versions mentioned there as a priority for other crates in its build chain, even if those crates can potentially reference more recent versions of the same crates. But I need to confirm that somehow. It could just be coincidence.

I also don't know what to expect for consumers of hREA building higher-order Holochain-based functionality on top of it. I have still not committed `Cargo.lock` files to either *DNA Auth Resolver* or *Agent Registration* projects; and these modules are *also* intended to be used as combinations of binaries and libraries in other projects, so it seems logical that they should be following the same crate configuration patterns as hREA (and currently, the absence of lockfiles means they are not).

Not quite sure what my question is in trying to explain this, I guess I'm just trying to solicit feedback on any places where this build and deployment setup can be improved so that there's less friction in working compatible code across an ecosystem.

pospi:
Ah

> Users dependent on the library will not inspect the library’s Cargo.lock (even if it exists).

So yes, should put library files in those other modules

pospi:
Just noting here that all integration tests are passing again, so we can probably go ahead with this.

I'd also like to see checks configured so that feature branches can't be merged into `sprout` without the integration test suite passing :slightly_smiling_face:

Connoropolous:
closing in favor of #267 since the git histories became diverged in a weird way I had trouble with

PR #205, opened 2022-02-27
fix the missing entry def for units

Connoropolous:
couldn't call `createUnit`


Connoropolous:
closing in favor of the other one

Issue #196, opened 2022-02-23
Implement conflict resolution behaviours for records


pospi:
There needs to be some means of recovering from an update conflict in the event that two concurrent edits are made to the same record during a network fork.

Some work has already been done [at a protocol level](https://lab.allmende.io/valueflows/vf-schemas/vf-graphql/-/issues/88) to afford all necessary metadata to handle this, but the implementation remains unfinished. #40 is also related, as author metadata needs to be included with records in order for revision conflicts to be understandable and manageable by humans.

Core logic for this would be best placed in the `hdk_records` crate. Specifically-

- [ ] `get_latest_header_hash` should be deprecated in favour of some alternative logic which returns an error if there are multiple live `updates` for the same entry. The error must include structured data containing the `HeaderHash`es of the conflicting revisions, such that clients can use that information to recover.
- #351
- [ ] `update_record` should also be changed to throw an error if the specified version is not the latest live `HeaderHash` for the associated `Entry`. This would prevent edits to old revisions under normal network conditions.
- [ ] some means of recovering from a conflict must also be implemented. This may be a case of the author(s) of a conflicting version requesting an update of the record to the same content as other divergent branches (in which case, Holochain's DHT structure might automatically re-merge without additional custom logic); or, if not possible, it might require implementing some special API call to mark one branch as superceded by another. If the latter, this might imply new mutations be added to the ValueFlow GraphQL spec.

Once Holochain core has implemented it we will probably need to address `EntryDhtStatus::Conflict` and potentially change the way `update_record` functions to use new host API calls.

A few moving parts to be implemented & enabled, in addition to the core logic:

- [ ] plumb errors through to the GraphQL client, such that records in conflict return an error that will be correctly interpreted by GraphQL *for that specific record* - i.e. any other non-conflicting records in the same request should still be returned successfully; the error cannot cause the whole query to fail.
- [ ] add an integration test to prove that the conflict detection behaviour of `update_entry` works under normal conditions, and old revisions can't be edited.
- [ ] add test for network forks, asserting correct error responses in the case of a conflict (after the network partition is resolved) and showcasing how to recover from a conflict.

PR #209, opened 2022-03-01
fixes #208 - update multiple times

### LINKED ISSUE #208: creating a 'raise' or 'lower' economic event for the same resource more than once doesn't produce update

Connoropolous:
fixes #208 

full background is in the commit message itself, but also here:

the original code didn't suit this context, and it was adapted from h-be/Acorn or hdk_crud.
It didn't suit because in Acorn HeaderHash is used as "id" where as in holo-rea EntryHash is used as "id"
hdk_crud creates a flat update tree, where all updates reference the original header
holo-rea is supposed to create a branching Update tree, which this now allows for.

However, this implementation is still very naive, and assumes that in fact the Update tree
mostly just nests children in one sequence, more like an linked list than a tree. This won't hold
true in practice, as we're dealing in offline-friendly distributed systems. Solving this properly
relates to issue https://github.com/holo-rea/holo-rea/issues/196 conflict resolution

### DIFF
--- a/lib/hdk_records/src/record_helpers.rs
+++ b/lib/hdk_records/src/record_helpers.rs
@@ -40,6 +40,10 @@ fn get_header_hash(shh: element::SignedHeaderHashed) -> HeaderHash {
 ///
 /// Useful in coordinating updates between different entry types.
 ///
+/// NOTE: this is a very naive recursive algorithm that basically assumes full network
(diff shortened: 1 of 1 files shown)


pospi (approved):
This is great! Should resolve bugs resulting from changes to the update logic; and should work well as-is for linear revisions to the same update. Though, I expect we will eventually want those to be detected as conflicts when #196 is implemented?

PR #224, opened 2022-03-23
fixups in vf-graphql-holochain

Connoropolous:
`agreement` was missing from `DNAIdMappings`

also each one seems to be optional, since it depends on the configuration

- [x] There are some outdated references to `hasMeasurement` in the vf-graphq-holochain module that need to be replaced with `hasSpecification`

### DIFF
--- a/modules/vf-graphql-holochain/mutations/index.ts
+++ b/modules/vf-graphql-holochain/mutations/index.ts
@@ -33,7 +33,7 @@ export type deleteHandler = (root: any, args: { revisionId: string }) => Promise
 export default (enabledVFModules: string[] = DEFAULT_VF_MODULES, dnaConfig: DNAIdMappings, conductorUri: string) => {
   const VFmodules = enabledVFModules || []
   const hasAgent = -1 !== VFmodules.indexOf("agent")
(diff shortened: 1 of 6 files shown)


pospi:
The missing `Agreement` cell is a good find, but fixed separately in eab164b0.

This is where the distinction to "VF modules" (which are defined as logical units of functionality in the specification) and "Holochain DNAs" (which are bundled differently based on other concerns) becomes important. The existing checks were actually correct as-is: hREA bundles the "measurement" datatypes (VF module) along with "knowledge" datatypes (VF module) in the "specification" DNA.

pospi:
I think that's right? @Connoropolous let me know if you were experiencing any actual bugs as a result of this discrepancy, but if you just made these changes based on an assumption then you can feel free to delete the branch. I'll leave it there until you review, just to be sure (:

Connoropolous:
Ooh. Thanks for clarifying. I thought I had solidified my understanding. 

Is the number of VF modules and their names equal to the files in vf-schemas? https://lab.allmende.io/valueflows/vf-schemas/vf-graphql/-/tree/sprout/lib/schemas

PR #228, opened 2022-03-24
fix #225: rename happs folder to dna_bundles

### LINKED ISSUE #225: rename top level folder `happs` to `dna_bundles` for clarity
since it contains dna definitions, and `happ` has a different meaning in the holochain ontology

Connoropolous:
fixes #225


pospi:
Sorta feels like the number of directories and their naming conventions are getting a bit out of hand now, so I cleaned this up a bit across the board.

PR #236, opened 2022-03-28
add the VfModule enum, use it throughout, and export it

### LINKED ISSUE #234: in `modules/vf-graphql-holochain`, make the identifiers of VF modules explicit in typescript as an enum instead of raw strings
I am picturing it as an enum where the values are strings

enum VFModule {
  Measurement,
  Agreement,
  Etc,
}

And then the `enabledVFmodules` option could be set to a type `VFModule[]`

That way typescript could help constrain and validate inputs as well as provide autocomplete.

Other places in the code that currently reference the magic module strings could be replaced with imports and code based references. 
Hopefully this would avoid others from similar confusion as I experienced with it.

Connoropolous:
@pospi I gave this a try, to test my understanding. please let me know if I'm getting this right. 

After my last misunderstanding, I'm committed to figuring out what VfModule really equates to lol :)

fixes #234

### DIFF
--- a/modules/vf-graphql-holochain/types.ts
+++ b/modules/vf-graphql-holochain/types.ts
@@ -1,11 +1,6 @@
 /**
  * base types for GraphQL query layer
  *
- * 
- * @see https://github.com/valueflows/vf-graphql/blob/master/schemas/structs.gql
- *
- * @package: HoloREA prototype
- * @since:   2019-01-03
  * @package: HoloREA
  * @since:   2019-05-20
  */
--- a/modules/vf-graphql-holochain/types.ts
+++ b/modules/vf-graphql-holochain/types.ts
@@ -94,7 +94,6 @@ export enum VfModule {
   Proposal,
   Recipe,
   Scenario,
-  Util,
 }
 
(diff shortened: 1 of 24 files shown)


modules/vf-graphql-holochain/types.ts
```diff
@@ -1,6 +1,7 @@
 /**
  * base types for GraphQL query layer
  *
+ * 
  * @see https://github.com/valueflows/vf-graphql/blob/master/schemas/structs.gql
```
Connoropolous: @pospi I added the above line just so that I could ping you to ask about what this link should instead be? the one on line 5

pospi: Looks old to me, I took it out in 8675452c 'cos the comment you have lower down in regard to the actual enum seems more useful.

pospi (approved):
Looks great! Consider your understanding checked ;)

pospi:
Ah, followup fix in 6c4719c5 - these need to be defined as a string enum or they'll result in an empty schema when passed into `@valueflows/vf-graphql-holochain`, since it's a pure JS modules which string matches based on the schema filenames.

Issue #216, opened 2022-03-11
ensure all integration tests passing or skipped


Connoropolous:
It's looking like the "Planning" suite is the remaining one with indexing configuration related issues #252 , #254, and #255.

```
343 passed
13 failed
...........................................................................................................
...........................................................................................................
.........x.x.......x.....xx..................................................xxx.........x......x..
xx........................................x
```

- [x] #263
- [x] #257
- [x] #247
- [x] `test/economic-event` has 13/13 assertions passing
- [x] #253
- [x] #246
- [x] #248
- [x] #249
- [x] #254
- [x] #255
- [x] #252
- [x] #251
- [x] #256
- [x] #250

The below task is not the only sub-task, just one of many possible ones.
- [x] #242

Connoropolous:
@pospi I am trying out this library tap-dot instead of `faucet` for the test out put since I am finding the faucet output very difficult to get a read on. 

https://github.com/scottcorgan/tap-dot

It is giving me a nice output like this: 
<img width="442" alt="Screen Shot 2022-04-11 at 7 56 47 PM" src="https://user-images.githubusercontent.com/1409121/162870323-0bb13335-462c-44f1-b479-83d76aa3c30a.png">

Before any other progress on the tests is made, how can we organize the tests, such that perhaps only the ones that are updated, and the ones that are passing, are running?

The problem right now is that there's no way for me to tell which ones are passing and which ones aren't. Is there some solution we can come up with for this?

pospi:
> The problem right now is that there's no way for me to tell which ones are passing and which ones aren't. Is there some solution we can come up with for this?

I am still working through them one directory at a time, and just running the commands piecemeal for now. We could create a completion matrix for them if you think it's too cumbersome to proceed in that way?

pospi:
Some notes on what typically needs updating in each test script:

- player initialisation logic changes need to use the new `buildConfig` and `buildPlayer` API format (doing these will actually get you into the test logic proper)
- you need to use the mocking methods from `init.js` to mock identifiers in order for the backend to accept them. There is `mockAddress()` for `EntryHash`-based IDs; `mockAgentId()` for `AgentPubKey`-based IDs (which will largely go away after #219); and `mockldentifier()` for string-based IDs (such as `UnitId`). The default for these methods is to generate identifiers as strings for passing to the GraphQL layer. Tests using the low-level `call` API should pass `true` as an argument to generate raw `Buffer` objects appropriate for passing directly to Holochain.
[...]

PR #245, opened 2022-04-12
usability goals for integration tests PR

### LINKED ISSUE #257: usability goals for integration tests
- [ ] develop a way to see which test file is currently being run by `tape`, clearly marked out in the TAP output
- [ ] make tests more isolated from each other so that some can continue even if others fail
- [ ] use `tap-dot` or something similar to provide a better summary at the end of the test running about what happened, and which tests passed and which failed

Connoropolous:
Trying to fix #257 
This is preliminary work towards #216.

Goal of this PR is to make the test output, and runners, more usable, and clear


test/economic-resource/resource_logic.js
```diff
@@ -20,6 +20,17 @@ const testEventProps = {
 runner.registerScenario('EconomicResource & EconomicEvent record interactions', async (s, t) => {
   const { cells: [observation, specification] } = await buildPlayer(s, config, ['observation', 'specification'])
 
+  // HACK: this prevents a flaky issue
```
pospi: Have you reinstalled modules since encountering this issue? After bfde90bd we should no longer be hitting these intermittent timeouts. Unless you have a *really* slow machine... I guess increasing the timeout again is a possibility. But I would try ensuring you have the patch active and see if it resolves things.

pospi:
Glad to have found the recursion fix- yeah, there may occasionally be `EconomicEvents` logged against an `EconomicResource` which don't actually result in updating the state of the resource (eg. during stocktake if the discovered inventory count is as expected, you might not change anything but still log an `action: raise` with a quantity of 0 just to state that the observation was made and double-checked).

pospi:
A couple of things about the integration test shell script-

I don't see it as terribly worthwhile to add, especially given that more test folders will be added in future. It's just another file to keep in sync with other parts of the repo and the only real benefit is printing out lines about which tests are executing. Once they're all passing we will expect the whole suite to run in one command and TAP output will tell us everything we need to know, so it seems like an unnecessary addition to me.

Also to note that the test numbers you're seeing are in no way accurate. The ones with very low assertion counts are just failing before execution begins and so the runner can't pick up anything following where the script crashes. Once we make API updates for the latest Tryorama they will proceed past the initialisation and you will start to see much higher numbers in the TAP output.

Connoropolous:
> Glad to have found the recursion fix- yeah, there may occasionally be `EconomicEvents` logged against an `EconomicResource` which don't actually result in updating the state of the resource (eg. during stocktake if the discovered inventory count is as expected, you might not change anything but still log an `action: raise` with a quantity of 0 just to state that the observation was made and double-checked).

Yeah, I've gathered that now. Makes sense.

test/economic-resource/resource_logic.js
Connoropolous: It's not a timeout, it seems like a race condition.

Connoropolous:
There are two objectives I have: 

- make it easy to run a specific, individual test file (my script doesn't accomplish this really yet. recommendations for how to do that?)
- make it easy to see which results are associated with which test file. The TAP output seems not useful at all for that, even with tap-dot

Connoropolous:
I hotfixed in to `sprout` branch the `get_latest_header_hash` fix, which fixes the Economic Resource tests, so now this PRs only diff is related to test running

test/economic-resource/resource_logic.js
Connoropolous: Ah, it manifested in an unexpected way. You're right that those things were related. It didn't seem to me like they could be. I would remove these again

Connoropolous:
I've deleted the script, and for now, reduced this PR to a change from `faucet` to `tap-dot` to achieve the intended goal of receiving a better summary of things at the end of a test run

pospi (approved):

Issue #219, opened 2022-03-15
Implement REA Agent


fosterlynn:
### MVP (simplest implementation)

- [x] Author a storage zome which allows any Holochain agent to store a blob of (`name` (string), `note` (string), `image` (string), `primaryLocation` (string), `type` (string)) and get back an `AgentAddress` pointing to this data. You can minimally create an `vf_attributes_hdk::AgentAddress` via `AgentAddress::new(dna_info()?.dna_hash, entry_hash)`
- [x] create `zome` `lib` and `storage` folders/crates under `rea_agent` and implement according to set patterns as can be seen in `rea_agreement` and elsewhere.
- [x] add those 3 crates into the top level Cargo.toml workspace definition
- [x] Wire it up to the query resolvers in `modules/vf-graphql-holochain/queries/agent.ts`
- [x] Implement a new agent mutation handler in `modules/vf-graphql-holochain/mutations/agent.ts` for `createPerson`
- [x] Include this zome in the DNA manifest in `happs/agent/dna.yaml`
- [x] Implement something separate for `Organization`, perhaps just a separate API endpoint which creates entries with a different "type" field. Wire this up to `createOrganization`.
- [x] #300
- [x] #311
- [x] #303 
- [x] #301 
- [x] #298

pospi:
## Production-capable (minimum viable for ongoing usage)

What's missing from the simple MVP above is validation / ownership of profiles; and an ability to deal with updates to data.
- [ ] #297
- [ ] #133 
- [ ] #299
- [ ] add validation to prevent edits to agent profiles other than by the agent who created them
- [ ] implement a means for agents to attest that another agent key belongs to them, adding the new key to the list of keys allowed to manage an existing profile. Same can also be used to whitelist agents able to manage an organization's details.

Just an outline for now- we can refine this and split tasks out to separate issues as needed.

pospi:
RE geolocation- it looks like there are two older efforts that might be able to be resurrected.

https://github.com/holochain-open-dev/geolocation
https://github.com/vanarchist/holochain-point-of-interest

Both would suffice- the linkage between records should be weak. Locations need to be identifiable by some unique ID, which we can then associate with other record fields (eg. `Agent.primaryLocation`). At some point we will probably also want locations to be able to be updated without that ID changing; because it's not really possible to easily track down references and change them.

[...]

Issue #214, opened 2022-03-11
tests for `hdk_records` using MockHDK


Connoropolous:
tests for `hdk_records` using [new mocking framework](https://forum.holochain.org/t/unit-testing-zomes-with-mockhdk/5421) (and [example](https://github.com/holochain-open-dev/minimal-hdk-unit-test-example))

PR #258, opened 2022-04-19
Feature/40 record revisions metadata

pospi:
Made a first pass at metadata conversion helpers and am curious to get an early review from someone who has been in the weeds with the DHT before to see if there are any glaring mistakes.

Would also really appreciate pointers or stubbing out of how to test this with the HDK mocking framework as I haven't done that before and it would take me much longer to find my feet with it (:


lib/hdk_records/src/metadata_helpers.rs
```diff
@@ -0,0 +1,186 @@
...
+use crate::{ RecordAPIResult, DataIntegrityError };
+
+/// Metadata for a specific revision of a record, serializable for external transmission
+///
+#[derive(Clone, Serialize, Deserialize, SerializedBytes, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct RevisionMeta {
+    pub id: HeaderHash,
+    pub time: Timestamp,
+    pub author_id: AgentPubKey,
```
Connoropolous: is author_id as a property name part of the Valueflows spec?

Connoropolous (commented):
the pattern looks broadly sensible to me. I can do something with the MockHDK soon here, yeah

lib/hdk_records/src/metadata_helpers.rs
pospi: It is, but I take your point- there's no reason the two have to marry and actually this one doesn't anyway (it will have to be translated through a `whois` query against the REA Agent DNA in the GraphQL resolvers in order to be turned into an `AgentId` once #219 has been implemented). What would make more sense as idiomatically Holochain? `agent_pubkey`?

Connoropolous: Yeah something like... agent_pub_key to match the casing 👍

Connoropolous:
so @pospi you mean starting in on #214 would be helpful here?

pospi:
@Connoropolous have pushed a first run at this. If CI passes, we should merge. Otherwise we should prioritise finishing off before continuing on #340 as the changes are quite extensive and I expect there will be significant merge conflicts if we attempt both at once.

pospi:
Still to do: needs some tests in `core-architecture` to assert that the data is returned and that it changes after an update. There may also be some work to do in the GraphQL layer but I'm pretty sure both `id` and `time` will be passed through without intervention. Possibly the `history` module needs to be enabled in the VF-GraphQL modules whitelist as well.

Anyway, nearly there I think! Signing off for the weekend (:

Connoropolous:
nice! yes I see CI passing and I did a basic review of the code and saw no issues

PR #262, opened 2022-04-25
#219 -> Wire up REA Agent storage DNA

### LINKED ISSUE #219: Implement REA Agent

pospi:
Includes revising data storage architecture to deal with an extra layer of indirection wrapped around REA's `Agent`, such that `Organization`, `EcologicalAgent`, `Group` etc are representable; and that all agent types (including the singular `Person`) can be associated with custom metadata and storage for broadcasting profiles within the collaboration space.

To implement #219.


weswalla:
@pospi noted about `AddressibleIdentifier`

weswalla:
@Connoropolous @pospi create and update are working, but for some reason delete isn't. calling `delete_agent` spits out the following error: `"Could not convert entry to requested type\"`. I tried digging in to the functions and it seems like the error is coming from: https://github.com/h-REA/hREA/blob/9376f4ccccd71db1a94937e8f231748b6671404b/lib/hdk_records/src/entry_helpers.rs#L39 
but I'm not sure why there is an error when trying to convert from the serialized bytes to `EntryData`, as there don't seem to be any major differences from other `EntryData` types

pospi:
You'll need to use `EntryStorage` with that, which is a wrapper struct containing the identifier info in addition to the `EntryData`. It's the difference between what actually gets stored in the DHT vs what is meaningfully "record data" within the application.

Connoropolous (approved):
This is great start. This is big enough that I am going to merge this, and clearly map and track the remaining scopes of work for this

Issue #221, opened 2022-03-21
Abstractions for list output handling


Connoropolous:
Abstract away list output handling into `hdk_records` (or potentially a new library crate) so that each individual zome can do things less verbosely.

pospi:
It's now added inline by the `index_zome` proc macro. Since it's only used at the zome API boundary, this seemed like a sensible architecture. Close if you agree (:

Connoropolous:
since it follows with the other conventions, happy to roll with it for now 👍

Connoropolous:
or should it be closed when #268 is merged I guess

pospi:
There is a little more to do in addition to that, need to finalize the `Cursor` handling & test pagination

pospi:
Actually, fine to close this. #222 tracks cursor-related stuff.

Issue #84, opened 2019-11-07
Implement "read all" GraphQL query endpoints


pospi:
This will require changes to the `vf-graphql` spec, since none of these endpoints are parameterised yet. This also means thinking about the best way of doing pagination in distributed environments that is also compatible with traditional centralised systems.

See Column E here for [completions status](https://docs.google.com/spreadsheets/d/1bbp896cA6rc9A3OTCOfIn0DMYL-EoulXG2W90zG6oWw/edit#gid=0)

Most immediately:
- [x] #268 
- [x] #320 

Ongoing / later:
- [x] #222
- [ ] #85
- [x] #221
- [ ] #220

pospi:
Updated resources on anchor pattern:

https://github.com/holochain/anchors
https://github.com/holochain/holochain_anchors

Examples:

https://github.com/holochain/anchors/blob/master/zomes/motorcycles
https://github.com/holochain/anchors/blob/master/zomes/motorcycles/code/Cargo.toml
https://github.com/holochain/anchors/blob/master/zomes/motorcycles/code/src/lib.rs

pospi:
For an example of the super-basic version of this, see https://github.com/holo-rea/holo-rea/pull/129

pospi:
Note that #85 should be finalised for `EconomicEvent` and `EconomicResource` before implementing the other endpoints, to avoid multiple refactorings / edits.

pospi:
I think it's also worth rearchitecting this slightly in the process of finalising #85, before proceeding here.

There's redundant data split between zomes due to the "list all" API endpoints currently being present within the CRUD zomes. This results in `hdk_uuid_types` anchor entries being stored multiple times- once in the CRUD zome for "read all", and once in the indexing zomes for anchoring against other relationships. If we were to move the "read all" endpoints into the indexing zomes as well, the DHT data structures in the CRUD zomes would be much cleaner&mdash; they'd only need to store the actual entry data rather than link structures to point to them & allow iteration. The CRUD zome WASMs would also be quite a bit smaller because none of the indexing helpers would need to be compiled in.

pospi:
> There's redundant data split between zomes due to the "list all" API endpoints currently being present within the CRUD zomes. This results in `hdk_uuid_types` anchor entries being stored multiple times[..]

These [concerns](https://github.com/h-REA/hREA/issues/84#issuecomment-1078743966) are addressed in #268.

Connoropolous:
[...]

Issue #220, opened 2022-03-21
Time-ordered listing & indexing


Connoropolous:
Implement reliable, DHTops-friendly means of retrieving a time-ordered list of records, and order records by creation time. (suggest [this lib](https://github.com/holochain-open-dev/holochain-time-index))

- #268 
- #331

pospi:
A more complete set of requirements for this, given a page size of N:

- *All* query APIs must return records in a consistent sort order, no matter the query parameters provided
- Different ordering will need to be done independently of the query parameters

(and per #85, but should be considered now)-

- Pagination must work as expected (starting record given by its ID and able to read N records forward)
- Paginating should return N records or fewer (if difficult it can return >N, but probably we can do this simply by truncating results)

Connoropolous:
I've moved the details from the last comment  above over to the relevant issues
#85 for pagination
#331 for ordered data for queries

with that I'm going to close this one, it feels redundant and dilutes the clarity

PR #268, opened 2022-05-22
#84 -> work area 1: Read All endpoints

pospi:
Initial cleanup pass for #221; refactors list indexing to place responsibility of "list all" data structures in index zomes alongside other queries.

Just wanted to put this draft up for @weswalla & @Connoropolous benefit as we move into the read API portion of the MMR work, since this is a significant architectural refactor.


pospi:
Ok, looks good now. Had some lingering bugs to track down but looks good after most recent commits. "List all" functionality is now fully the responsibility of index zomes, rather than CRUD zomes.

Connoropolous:
I merged sprout branch in, so that CI will run here now 👍

Connoropolous:
@pospi there is one assertion failing on ubuntu in CI, does that also occur for you locally?

lib/hdk_records/src/record_helpers.rs
```diff
@@ -164,13 +168,17 @@ pub fn create_record<I, R: Clone, B, C, E, S>(
...
-    let identity_address = create_entry_identity(&entry_def_id, &identity)?;
+    create_entry_identity(
+        indexing_zome_name_from_config,
+        &entry_def_id, &identity,
+    )?;
 
     // link the identifier to the actual entry
     // :TODO: this isn't needed for reading anymore, but might be worthwhile retaining for legibility in the DHT. Needs consideration as to DHT size bloat tradeoff.
-    create_link(identity_address, entry_hash, LinkTag::new(crate::identifiers::RECORD_INITIAL_ENTRY_LINK_TAG))?;
+    // create_link(identity_address, entry_hash, LinkTag::new(crate::identifiers::RECORD_INITIAL_ENTRY_LINK_TAG))?;
```
Connoropolous: what's the reason for commenting it out, vs deleting it?

Connoropolous (commented):
I have scanned the code, but most importantly at the moment the integration tests suggest that the response for the read_all_* queries are coming back empty, so I assume things aren't quite fully functionally integrated yet

pospi:
The core of this issue is now fixed- the remaining errors in that test are due to list APIs returning duplicate results.

However since we are pursuing #220 I wonder whether it is a better course of action to fold that work into this PR and have the switchover of DHT indexing structures resolve the duplication issue as a side-effect?

Connoropolous:
@pospi cool. Where do the duplicate results come from?

pospi:
(for completeness) duplicate link writes between the indexing entries.

lib/hdk_time_indexing/src/lib.rs
```diff
@@ -0,0 +1,86 @@
...
+    Year,
+    Month,
+    Day,    // :NOTE: coarsest granularity of created indexes
+    Hour,
+    Minute,
+    Second,
+}
+
+// Parse configuration & setup library constants
+lazy_static! {
```
Connoropolous, line 81: I don't know if it matters much, but instead of lazy_static I know that the holochain team uses OnceCell now, and it looks like it's likely to get moved into `std` too

lib/hdk_time_indexing/src/writing.rs
```diff
@@ -0,0 +1,97 @@
...
+fn link_if_not_linked(origin_hash: EntryHash, dest_hash: EntryHash, link_tag: LinkTag) -> TimeIndexResult<()> {
+    if false == get_links(origin_hash.to_owned(), Some(link_tag.to_owned()))?
+        .iter().any(|l| { EntryHash::from(l.target.to_owned()) == dest_hash })
+    {
+        create_link(
+            dest_hash.to_owned(),
+            origin_hash.to_owned(),
+            HdkLinkType::Any,
+            link_tag,
+        ).map_err(|_e| { TimeIndexingError::EntryNotFound(origin_hash.to_owned()) })?;
```
Connoropolous, line 93: why this line? I'm not sure it reflects real holochain behaviour, does it?

lib/hdk_semantic_indexes/zome/src/lib.rs
```diff
@@ -100,6 +105,50 @@ pub fn query_index<'a, T, O, C, F, A, S, I, J, E>(
     Ok(entries)
 }
 
+/// Given a type of entry, returns a Vec of *all* records of that entry registered
```
Connoropolous: won't be "all" if the caller constrains the request by time frame

Connoropolous (commented):
this is just my first read through. I was moderate thorough. I haven't done anything hands on

lib/hdk_time_indexing/src/lib.rs
pospi: ah, well that's good to know. Advice might be the outcome of https://github.com/holochain/holochain/issues/1444 then

lib/hdk_time_indexing/src/writing.rs
pospi: you mean the error mapping? (sorry, it's not clear exactly which line you mean in the mobile Github UI :S)

Connoropolous: The error mapping yeah

Connoropolous (commented):
I tried to check out this branch and run it but there were a few Rust compilation failures. I could figure out three of them, but not why `get_links_for_time_span` is being called but is entirely undefined in the codebase. I'd happily check out this branch again and do some looking into it if I can have it compile.

pospi:
My bad, that method is being updated to not target the `hc_time_index` module and I hadn't pushed. Have just put through a commit to comment it out until the logic at least compiles.

lib/hdk_semantic_indexes/zome/src/lib.rs
pospi: will update comment header when I update the method (:

lib/hdk_time_indexing/src/writing.rs
pospi: Gotcha. Well, "real" behaviour- probably not?
I'm moving to a pattern where records have to be committed to the local DHT before any indexes can be created for them. Calls to `create_link` fail if the entry being linked to doesn't exist, and I wanted to catch this as the library's own error type so that the reason for erroring is a bit clearer to the caller. If you check `append_to_time_index` you'll see that it ensures the pointer entry being linked to exists before populating the index data structures.

Connoropolous:
@pospi I don't have additional time today past my solving for the logging issue, which I've accomplished (see #305 ) but I did also push a fix on this branch for an issue where if a test fails, `hc sandbox` could leave spawned holochain and lair-keystore processes running orphaned on the machine.

pospi:
Ah, great! Hectic change, thanks for taking the time (:

`ps aux | grep holochain` found like 12 of those I still had running!

pospi:
First pass of this seems to be working! There is now a simple "complete traversal" method that is passing tests for the event & resource list APIs.
Ready for review- we can build up the more complex algorithms from this foundation. I am keeping a list of TODO items separately that I will probably log in a separate repo for the new `hdk_time_indexing` crate. Once we are happy with an MVP I think it will be best to split it out into its own repo and seek some systems programmers to take over its continuation :slightly_smiling_face:

lib/hdk_semantic_indexes/zome/src/lib.rs
```diff
@@ -100,6 +110,48 @@ pub fn query_index<'a, T, O, C, F, A, S, I, J, E>(
...
+    _limit: usize,
+) -> RecordAPIResult<Vec<RecordAPIResult<T>>>
+    where T: serde::de::DeserializeOwned + std::fmt::Debug,
+        B: DnaAddressable<EntryHash> + TryFrom<SerializedBytes, Error = SerializedBytesError>,
+        I: AsRef<str> + std::fmt::Display,
+        C: std::fmt::Debug,
+        SerializedBytes: TryInto<C, Error = SerializedBytesError> + TryInto<B, Error = SerializedBytesError>,
+        F: Fn(C) -> Option<String>,
+{
+    let linked_records = read_all_entry_hashes(index_name)
```
Connoropolous, line 138:
```suggestion
    // this algorithm is the 'make it work' current pass, pending the full implementation mentioned
    // in the TODO below, regarding efficiency and completeness
    let linked_records = read_all_entry_hashes(index_name)
```

Connoropolous (approved):
Fantastic. I see that you also added the missing indexes, which means that the only thing needed to make `processes`, `agents`, etc resolvers work is to wire up the resolvers, since you've moved the `read_all_x` endpoints into the index zomes. I ran the integration tests for the branch and they all passed, after making one small fix to from `vf_plan` to `plan` which was just a delayed merge thing

PR #280, opened 2022-06-03
Map zome generic params

### LINKED ISSUE #270: make `mapZomeFn` take generic parameters so that we have compiler guarantees at API boundary
- [ ] make the function accept generic arguments for the input and output type
- [ ] for each `mapZomeFn(...)` call import the relevant types and reference them

Not sure if this is the moment, but noting that I'd like to make `mapZomeFn` take generic parameters so that we can define both the expected input type and the expected output type. 
That would mean we'd be able to have type checking and type safety on this line where we write `(...).plan` and it's so opaque why there is a property called `plan` on this data structure. 

e.g. it might look like `mapZomeFn<InputType, OutputType>(...)`

_Originally posted by @Connoropolous in https://github.com/h-REA/hREA/pull/269#discussion_r880052534_

Related code: 
https://github.com/h-REA/hREA/blob/459c8c98e224a3330350a38ac6b0c9178dab88df/modules/vf-graphql-holochain/connection.ts#L266-L296

Instead of having the 'explicit type loss' we can retain type information where/when we have it, and write `<any, any>` in the worst case scenario of not knowing the types (which... should never happen)

weswalla:
The main typescript error feedback we are getting now has to do with a slight different in a type definition and the actual object that gets passed in.

for example, from the `economicResource` resolver: 
```
const readResourceSpecification = mapZomeFn<ReadParams, ResourceSpecificationResponse>(dnaConfig, conductorUri, 'specification', 'resource_specification', 'get_resource_specification')

...

conformsTo: async (record: EconomicResource): Promise<ResourceSpecification> => {
        return (await readResourceSpecification({ address: record.conformsTo})).resourceSpecification
      },
```

because we declare record as `EconomicResource` then ts thinks that record.conformsTo is of type `ResourceSpecification` but really at this layer it is an id, which we use to return a ResourceSpecification. Which makes me wonder how to properly type the `record` input, as it isn't quite the `EconomicResource` object, but the base object which contains the ids of all the linked objects.

### DIFF
--- a/modules/vf-graphql-holochain/mutations/economicEvent.ts
+++ b/modules/vf-graphql-holochain/mutations/economicEvent.ts
@@ -15,6 +15,7 @@ import {
   EconomicEventUpdateParams,
   EconomicEventResponse,
 } from '@valueflows/vf-graphql'
+import { HeaderHash } from '@holochain/client'
 
 export interface CreateArgs {
   event: EconomicEventCreateParams,
@@ -25,12 +26,16 @@ export type createHandler = (root: any, args: CreateArgs) => Promise<EconomicEve
 export interface UpdateArgs {
   event: EconomicEventUpdateParams,
 }
+
+export interface DeleteArgs {
+  revisionId: string,
+}
 export type updateHandler = (root: any, args: UpdateArgs) => Promise<EconomicEventResponse>
 
 export default (dnaConfig: DNAIdMappings, conductorUri: string) => {
-  const runCreate = mapZomeFn(dnaConfig, conductorUri, 'observation', 'economic_event', 'create_economic_event')
-  const runUpdate = mapZomeFn(dnaConfig, conductorUri, 'observation', 'economic_event', 'update_economic_event')
-  const runDelete = mapZomeFn(dnaConfig, conductorUri, 'observation', 'economic_event', 'delete_economic_event')
+  const runCreate = mapZomeFn<CreateArgs, EconomicEventResponse>(dnaConfig, conductorUri, 'observation', 'economic_event', 'create_economic_event')
+  const runUpdate = mapZomeFn<UpdateArgs, EconomicEventResponse>(dnaConfig, conductorUri, 'observation', 'economic_event', 'update_economic_event')
+  const runDelete = mapZomeFn<DeleteArgs, boolean>(dnaConfig, conductorUri, 'observation', 'economic_event', 'delete_economic_event')
 
   const createEconomicEvent: createHandler = async (root, args) => {
     return runCreate(args)
--- a/modules/vf-graphql-holochain/mutations/economicEvent.ts
+++ b/modules/vf-graphql-holochain/mutations/economicEvent.ts
@@ -5,7 +5,7 @@
  * @since:   2019-05-27
  */
 
-import { DNAIdMappings } from '../types'
+import { ByRevision, DNAIdMappings } from '../types'
 import { mapZomeFn } from '../connection'
 import { deleteHandler } from './'
 
@@ -27,15 +27,12 @@ export interface UpdateArgs {
   event: EconomicEventUpdateParams,
 }
 
-export interface DeleteArgs {
-  revisionId: string,
-}
 export type updateHandler = (root: any, args: UpdateArgs) => Promise<EconomicEventResponse>
 
 export default (dnaConfig: DNAIdMappings, conductorUri: string) => {
   const runCreate = mapZomeFn<CreateArgs, EconomicEventResponse>(dnaConfig, conductorUri, 'observation', 'economic_event', 'create_economic_event')
   const runUpdate = mapZomeFn<UpdateArgs, EconomicEventResponse>(dnaConfig, conductorUri, 'observation', 'economic_event', 'update_economic_event')
-  const runDelete = mapZomeFn<DeleteArgs, boolean>(dnaConfig, conductorUri, 'observation', 'economic_event', 'delete_economic_event')
+  const runDelete = mapZomeFn<ByRevision, boolean>(dnaConfig, conductorUri, 'observation', 'economic_event', 'delete_economic_event')
 
   const createEconomicEvent: createHandler = async (root, args) => {
     return runCreate(args)
--- a/modules/vf-graphql-holochain/mutations/processSpecification.ts
+++ b/modules/vf-graphql-holochain/mutations/processSpecification.ts
@@ -5,7 +5,7 @@
  * @since:   2019-09-12
  */
 
-import { DNAIdMappings } from '../types'
+import { ByRevision, DNAIdMappings } from '../types'
 import { mapZomeFn } from '../connection'
 import { deleteHandler } from './'
 
@@ -26,24 +26,20 @@ export interface UpdateArgs {
 export type updateHandler = (root: any, args: UpdateArgs) => Promise<ProcessSpecificationResponse>
 
 export default (dnaConfig: DNAIdMappings, conductorUri: string) => {
-  const runCreate = mapZomeFn(dnaConfig, conductorUri, 'specification', 'process_specification', 'create_process_specification')
-  const runUpdate = mapZomeFn(dnaConfig, conductorUri, 'specification', 'process_specification', 'update_process_specification')
-  const runDelete = mapZomeFn(dnaConfig, conductorUri, 'specification', 'process_specification', 'delete_process_specification')
+  const runCreate = mapZomeFn<CreateArgs, ProcessSpecificationResponse>(dnaConfig, conductorUri, 'specification', 'process_specification', 'create_process_specification')
+  const runUpdate = mapZomeFn<UpdateArgs, ProcessSpecificationResponse>(dnaConfig, conductorUri, 'specification', 'process_specification', 'update_process_specification')
+  const runDelete = mapZomeFn<ByRevision, boolean>(dnaConfig, conductorUri, 'specification', 'process_specification', 'delete_process_specification')
 
   const createProcessSpecification: createHandler = async (root, args) => {
-    return runCreate({
-      process_specification: args.processSpecification
-    })
+    return runCreate(args)
   }
 
   const updateProcessSpecification: updateHandler = async (root, args) => {
-    return runUpdate({
-      process_specification: args.processSpecification
-    })
+    return runUpdate(args)
   }
 
   const deleteProcessSpecification: deleteHandler = async (root, args) => {
-    return runDelete({ address: args.revisionId })
+    return runDelete(args)
   }
 
   return {
--- a/modules/vf-graphql-holochain/mutations/proposedIntent.ts
+++ b/modules/vf-graphql-holochain/mutations/proposedIntent.ts
@@ -12,7 +12,6 @@ import { deleteHandler } from './'
 import {
   ProposedIntentResponse,
 } from '@valueflows/vf-graphql'
-import { CreateCloneCellRequest } from '@holochain/client'
 
 export interface CreateParams {
   proposedIntent: CreateRequest,
--- a/modules/vf-graphql-holochain/mutations/proposedIntent.ts
+++ b/modules/vf-graphql-holochain/mutations/proposedIntent.ts
@@ -5,7 +5,7 @@
  * @since:   2019-09-12
  */
 
-import { DNAIdMappings } from '../types'
+import { ByRevision, DNAIdMappings, IntentAddress, ProposalAddress } from '../types'
 import { mapZomeFn } from '../connection'
 import { deleteHandler } from './'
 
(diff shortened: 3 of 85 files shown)


modules/vf-graphql-holochain/queries/intent.ts
```diff
@@ -5,20 +5,20 @@
...
 
 // :TODO: how to inject DNA identifier?
 export default (dnaConfig: DNAIdMappings, conductorUri: string) => {
-  const readRecord = mapZomeFn(dnaConfig, conductorUri, 'planning', 'intent', 'get_intent')
+  const readRecord = mapZomeFn<ReadParams, IntentResponse>(dnaConfig, conductorUri, 'planning', 'intent', 'get_intent')
 
   return {
     intent: async (root, args): Promise<Intent> => {
-      return (await (await readRecord)({ address: args.id })).intent
+      return (await (await readRecord)(args)).intent
```
weswalla: whats up with this syntax?

Connoropolous: that's look like an auto-format of a little syntax mistake, but that still worked in practice. could delete the extra (await ) but leave the `readRecord` it wraps

modules/vf-graphql-holochain/queries/action.ts
```diff
@@ -5,24 +5,24 @@
...
 
 import {
   Action,
 } from '@valueflows/vf-graphql'
 
 export default (dnaConfig: DNAIdMappings, conductorUri: string) => {
-  const read = mapZomeFn(dnaConfig, conductorUri, 'specification', 'action', 'get_action')
-  const readAll = mapZomeFn(dnaConfig, conductorUri, 'specification', 'action', 'get_all_actions')
+  const read = mapZomeFn<ReadParams, Action>(dnaConfig, conductorUri, 'specification', 'action', 'get_action')
+  // const readAll = mapZomeFn<null, Action[]>(dnaConfig, conductorUri, 'specification', 'action', 'get_all_actions')
```
Connoropolous: oh sorry, this one is an exception to the others, and exists. plz uncomment. It is different because it doesn't use semantic_index, it just returns a flat pre-defined list that's defined in the code, as opposed to querying holochain

modules/vf-graphql-holochain/queries/action.ts
```diff
@@ -5,24 +5,24 @@
...
     action: async (root, args): Promise<Action> => {
       return read(args)
     },
 
-    actions: async (root, args): Promise<Action> => {
-      return readAll(null)
-    },
+    // actions: async (root, args): Promise<Action[]> => {
+    //   return readAll(null)
+    // },
```
Connoropolous: same as above, just plz uncomment

modules/vf-graphql-holochain/queries/agreement.ts
```diff
@@ -5,19 +5,20 @@
...
 } from '@valueflows/vf-graphql'
 
 export default (dnaConfig: DNAIdMappings, conductorUri: string) => {
-  const readRecord = mapZomeFn(dnaConfig, conductorUri, 'agreement', 'agreement', 'get_agreement')
+  const readRecord = mapZomeFn<ReadParams, AgreementResponse>(dnaConfig, conductorUri, 'agreement', 'agreement', 'get_agreement')
 
   return {
-    agreement: async (root, { id }): Promise<Agreement> => {
-      return (await readRecord({ address: id })).agreement
+    // why does agreement query not return AgreementResponse like other queries?
```
Connoropolous: I'm not sure I follow your comment/question here. It appears to me to follow in the same pattern as the others

modules/vf-graphql-holochain/mutations/economicEvent.ts
```diff
@@ -25,12 +26,16 @@ export type createHandler = (root: any, args: CreateArgs) => Promise<EconomicEve
 export interface UpdateArgs {
   event: EconomicEventUpdateParams,
 }
+
+export interface DeleteArgs {
+  revisionId: string,
+}
```
Connoropolous: Can you delete this one and replace with imported `ByRevision` type?

modules/vf-graphql-holochain/resolvers/intent.ts
```diff
@@ -53,14 +58,14 @@ export default (enabledVFModules: VfModule[] = DEFAULT_VF_MODULES, dnaConfig: DN
       },
     } : {}),
     (hasObservation ? {
-      inputOf: async (record: Intent): Promise<Process[]> => {
+      inputOf: async (record: Intent): Promise<Process> => {
```
Connoropolous, line 61: nice fix

modules/vf-graphql-holochain/mutations/resourceSpecification.ts
```diff
@@ -26,24 +26,20 @@ export interface UpdateArgs {
...
-  const runDelete = mapZomeFn(dnaConfig, conductorUri, 'specification', 'resource_specification', 'delete_resource_specification')
+  const runCreate = mapZomeFn<CreateArgs, ResourceSpecificationResponse>(dnaConfig, conductorUri, 'specification', 'resource_specification', 'create_resource_specification')
+  const runUpdate = mapZomeFn<UpdateArgs, ResourceSpecificationResponse>(dnaConfig, conductorUri, 'specification', 'resource_specification', 'update_resource_specification')
+  const runDelete = mapZomeFn<ByRevision, boolean>(dnaConfig, conductorUri, 'specification', 'resource_specification', 'delete_resource_specification')
 
   const createResourceSpecification: createHandler = async (root, args) => {
-    return runCreate({
-      resource_specification: args.resourceSpecification
-    })
+    return runCreate(args)
```
Connoropolous, line 34: so there's a fix needed here. You did the right thing, but only half so. 

You need to go in to `rpc` folder of the `rea_resource_specification` zome, and add serde rename all camelCase for these two structs.

modules/vf-graphql-holochain/mutations/proposedTo.ts
```diff
@@ -5,26 +5,33 @@
...
 
 export default (dnaConfig: DNAIdMappings, conductorUri: string) => {
-  const runCreate = mapZomeFn(dnaConfig, conductorUri, 'proposal', 'proposed_to', 'create_proposed_to')
-  const runDelete = mapZomeFn(dnaConfig, conductorUri, 'proposal', 'proposed_to', 'delete_proposed_to')
+  const runCreate = mapZomeFn<CreateArgs, ProposedToResponse>(dnaConfig, conductorUri, 'proposal', 'proposed_to', 'create_proposed_to')
+  const runDelete = mapZomeFn<ByRevision, boolean>(dnaConfig, conductorUri, 'proposal', 'proposed_to', 'delete_proposed_to')
 
   const proposeTo: createHandler = async (root, args) => {
-    return runCreate({ proposed_to: args })
+    return runCreate(args)
```
Connoropolous: you also changed the input type. Make sure that serde rename all is right on the related zome struct

modules/vf-graphql-holochain/mutations/proposedIntent.ts
```diff
@@ -5,26 +5,33 @@
...
 
 export default (dnaConfig: DNAIdMappings, conductorUri: string) => {
-  const runCreate = mapZomeFn(dnaConfig, conductorUri, 'proposal', 'proposed_intent', 'create_proposed_intent')
-  const runDelete = mapZomeFn(dnaConfig, conductorUri, 'proposal', 'proposed_intent', 'delete_proposed_intent')
+  const runCreate = mapZomeFn<CreateArgs, ProposedIntentResponse>(dnaConfig, conductorUri, 'proposal', 'proposed_intent', 'create_proposed_intent')
+  const runDelete = mapZomeFn<ByRevision, boolean>(dnaConfig, conductorUri, 'proposal', 'proposed_intent', 'delete_proposed_intent')
 
   const proposeIntent: createHandler = async (root, args) => {
-    return runCreate({ proposed_intent: args })
+    return runCreate(args)
```
Connoropolous: check the zome type

modules/vf-graphql-holochain/mutations/processSpecification.ts
```diff
@@ -26,24 +26,20 @@ export interface UpdateArgs {
...
-  const runDelete = mapZomeFn(dnaConfig, conductorUri, 'specification', 'process_specification', 'delete_process_specification')
+  const runCreate = mapZomeFn<CreateArgs, ProcessSpecificationResponse>(dnaConfig, conductorUri, 'specification', 'process_specification', 'create_process_specification')
+  const runUpdate = mapZomeFn<UpdateArgs, ProcessSpecificationResponse>(dnaConfig, conductorUri, 'specification', 'process_specification', 'update_process_specification')
+  const runDelete = mapZomeFn<ByRevision, boolean>(dnaConfig, conductorUri, 'specification', 'process_specification', 'delete_process_specification')
 
   const createProcessSpecification: createHandler = async (root, args) => {
-    return runCreate({
-      process_specification: args.processSpecification
-    })
+    return runCreate(args)
```
Connoropolous, line 34: same as resourceSpecification, add serde rename all  to the Create and Update structs in `rpc` folder of `rea_process_specification` if you make this change. Making this change is the right thing to do

Connoropolous (changes requested):
nice! A few pointers on areas of completion for this. It became larger scope than I thought, because we caught rust side stuff crying for minor changes.

modules/vf-graphql-holochain/queries/agreement.ts
weswalla: I think I meant on the vf gql definition side of things: https://lab.allmende.io/valueflows/vf-schemas/vf-graphql/-/blob/0.9.x/lib/schemas/agreement.gql#L76

Connoropolous: ah, looks unintended

modules/vf-graphql-holochain/mutations/proposedIntent.ts
```diff
@@ -5,26 +5,33 @@
...
 
 export default (dnaConfig: DNAIdMappings, conductorUri: string) => {
-  const runCreate = mapZomeFn(dnaConfig, conductorUri, 'proposal', 'proposed_intent', 'create_proposed_intent')
-  const runDelete = mapZomeFn(dnaConfig, conductorUri, 'proposal', 'proposed_intent', 'delete_proposed_intent')
+  const runCreate = mapZomeFn<CreateArgs, ProposedIntentResponse>(dnaConfig, conductorUri, 'proposal', 'proposed_intent', 'create_proposed_intent')
+  const runDelete = mapZomeFn<ByRevision, boolean>(dnaConfig, conductorUri, 'proposal', 'proposed_intent', 'delete_proposed_intent')
 
   const proposeIntent: createHandler = async (root, args) => {
-    return runCreate({ proposed_intent: args })
+    return runCreate({ proposedIntent: args })
```
Connoropolous, line 31:
```suggestion
    return runCreate(args)
```
I think without reverting, this will be a bug. Add the typescript definition that's missing for `args` to the above `createHandler` type definition.

modules/vf-graphql-holochain/mutations/proposedTo.ts
```diff
@@ -5,26 +5,32 @@
...
 
 export default (dnaConfig: DNAIdMappings, conductorUri: string) => {
-  const runCreate = mapZomeFn(dnaConfig, conductorUri, 'proposal', 'proposed_to', 'create_proposed_to')
-  const runDelete = mapZomeFn(dnaConfig, conductorUri, 'proposal', 'proposed_to', 'delete_proposed_to')
+  const runCreate = mapZomeFn<CreateArgs, ProposedToResponse>(dnaConfig, conductorUri, 'proposal', 'proposed_to', 'create_proposed_to')
+  const runDelete = mapZomeFn<ByRevision, boolean>(dnaConfig, conductorUri, 'proposal', 'proposed_to', 'delete_proposed_to')
 
   const proposeTo: createHandler = async (root, args) => {
-    return runCreate({ proposed_to: args })
+    return runCreate({ proposedTo: args })
```
Connoropolous, line 30:
```suggestion
    return runCreate(args)
```
I think without reverting, this will be a bug. Add the typescript definition that's missing for `args` to the above `createHandler` type definition.

modules/vf-graphql-holochain/mutations/proposedIntent.ts
weswalla: I think my new changes are right, this is a bit weird because I think there is a bit of a discrepancy in input types for `create` mutations, if the mutation has a defined input type (any `CreateParams` object), vs multiple inputs and how gql handles this (I think?)

Connoropolous: ah ok. Let me review what you're saying about that.

modules/vf-graphql-holochain/mutations/proposedIntent.ts
```diff
@@ -5,26 +5,35 @@
...
 
-import { DNAIdMappings } from '../types'
+import { ByRevision, DNAIdMappings, IntentAddress, ProposalAddress } from '../types'
 import { mapZomeFn } from '../connection'
 import { deleteHandler } from './'
 
 import {
   ProposedIntentResponse,
 } from '@valueflows/vf-graphql'
+import { CreateCloneCellRequest } from '@holochain/client'
```
Connoropolous:
```suggestion
```

modules/vf-graphql-holochain/mutations/proposedIntent.ts
Connoropolous: Here are two resources we can build off: 
https://www.apollographql.com/tutorials/fullstack-quickstart/writing-query-resolvers
https://graphql.org/graphql-js/passing-arguments/

modules/vf-graphql-holochain/queries/agreement.ts
Connoropolous: as follow up, I logged this: https://lab.allmende.io/valueflows/vf-schemas/vf-graphql/-/issues/99

modules/vf-graphql-holochain/mutations/proposedTo.ts
Connoropolous: I retract, thanks for the clarification. In the future it would be good to clean this up. I've heard these proposal things are somewhat relatively new, but they broke from the convention a bit so they look out of place.

modules/vf-graphql-holochain/mutations/proposedIntent.ts
Connoropolous: You're correct, we could follow up on this eventually potentially to make it the same as the others.

Connoropolous (approved):
It looks done to me 👍

Connoropolous:
@weswalla since this is a branch of a branch, CI isn't running on it. Have you ran the tests?

weswalla:
tests now passing @Connoropolous

pospi:
> ts thinks that record.conformsTo is of type ResourceSpecification but really at this layer it is an id, which we use to return a ResourceSpecification. Which makes me wonder how to properly type the record input, as it isn't quite the EconomicResource object, but the base object which contains the ids of all the linked objects.

This is a good question. I think it is mostly worked around at the moment since Typescript doesn't do any runtime type-checking and so in the way our resolver accesses those fields it doesn't care & "just so happens" to work?

PR #286, opened 2022-06-08
refactored resolver composition to reflect new modules

### LINKED ISSUE #277: once vf-graphql has Vf Module changes, update the version and update the necessary code in vf-graphql-holochain
implement these changes: 
https://lab.allmende.io/valueflows/vf-schemas/vf-graphql/-/merge_requests/105

https://lab.allmende.io/valueflows/vf-schemas/vf-graphql/-/tree/module-breakdown-and-bridging/lib/schemas

- look for instances of `hasKnowledge` `hasPlanning` or `hasObservation` and replace those instances where necessary with new and more specific module names, as defined by the list of changes in the Gitlab pull request

weswalla:


modules/vf-graphql-holochain/types.ts
```diff
@@ -136,20 +136,27 @@ export enum VfModule {
 // default 'full suite' VF module set supported by Holo-REA
 
 export const DEFAULT_VF_MODULES = [
-  VfModule.Agent,
-  VfModule.Agreement,
+  // Specification DNA
```
pospi, line 139: I've been considering these groupings of zomes into DNAs "the default/MVP suite", but my hope is that eventually the particular composition will be meaningless and that 3rd-party app developers will be able to create their own arrangements of zomes and indexes without too much trouble.

pospi (approved):

modules/vf-graphql-holochain/types.ts
Connoropolous: Agreed!

Connoropolous: DevHub actually is a really good tool for mixing and matching.

pospi: DNAs? Or can it do per-zome remixing as well?

Connoropolous: Per zome as well.

PR #296, opened 2022-06-13
update tryorama, and @holochain/client to latest

### LINKED ISSUE #290: update tryorama to 0.5.4 - new syntax
this will help us unravel some of our little shims that we've added to fix tryorama ourselves

https://www.npmjs.com/package/@holochain/tryorama/v/0.5.4

Connoropolous:
closes #290 

- [x] updates @holochain/client to v0.4.0 (from just 6 days ago)
- [x] enables / uses ES `module` for `modules/vf-graphql-holochain`, and `test`
- [x] updated `tape` to version 5.5.3
- [x] eliminates tryorama shims


test/agreement/test_agreement_crud.js
```diff
@@ -1,28 +1,23 @@
-const {
-  buildConfig,
-  buildRunner,
-  buildPlayer,
-} = require('../init')
-
-const runner = buildRunner()
-
-const config = buildConfig()
+import test from "tape";
```
Connoropolous: I accidentally formatted this file

modules/vf-graphql-holochain/package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "@valueflows/vf-graphql-holochain",
   "version": "0.0.1-alpha.4",
+  "type": "module",
```
Connoropolous, line 4: This is a core change

test/init.js
```diff
@@ -5,236 +5,262 @@
...
-    dnas: dnas
-  }
-  await player.adminWs().installApp(installAppReq)
-  // must be enabled to be callable
-  const enabledAppResponse = await player.adminWs().enableApp({
-      installed_app_id: installAppReq.installed_app_id
-  })
-  if (enabledAppResponse.errors.length > 0) {
-      throw new Error(`Error - Failed to enable app: ${enabledAppResponse.errors}`)
+const buildPlayer = async (agentDNAs, graphQLAPIOptions) => {
```
Connoropolous, line 102: I also accidentally formatted this file. If I/we can establish a standard formatting method it'd be great.
I personally recommend `prettier`.

Anyways, this function `buildPlayer` is the only thing with logical changes.

modules/vf-graphql-holochain/mutations/agreement.ts
```diff
@@ -5,8 +5,8 @@
  * @since:   2020-06-19
  */
 
-import { ByRevision, DNAIdMappings } from '../types'
-import { mapZomeFn } from '../connection'
+import { ByRevision, DNAIdMappings } from '../types.js'
```
pospi, line 8: curious why you've added these... I shouldn't think adding extensions would be necessary?

test/init.js
pospi: I'm using https://standardjs.com/ - there is an `.eslintrc` in the `test/` directory with a config, and a `tslint.json` in one of the `modules/` (obviously, should be in both).

Your editor might not be picking these up- if not, feel free to add more dotfiles as needed.

Connoropolous: Ok I'll see if I can get those working. Using VS Code

modules/vf-graphql-holochain/mutations/agreement.ts
Connoropolous: I don't know why they were, but they were. Perhaps there is a configuration via which we could not use them.

modules/vf-graphql-holochain/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "outDir": "./build",
-    "module": "commonjs",
+    "module": "ES6",
```
Connoropolous, line 4: this is another core change

test/init.js
Connoropolous: I got it working. I formatted all files according to those rules

modules/vf-graphql-holochain/mutations/agreement.ts
Connoropolous: I guess because there's no transpilation. This is raw ES Modules we're using here now, since these are just .js files, not typescript. 

https://stackoverflow.com/a/63460038

test/init.js
Connoropolous: I had to update and tweak the .eslintrc file slightly.

Connoropolous:
@pospi I got the tests all passing on this branch, at least locally. Looking into CI

modules/vf-graphql-holochain/mutations/agreement.ts
pospi: But they're not, are they? There are still `type` & `interface` defs as well as function generics in these source files. I want to be sure we're not breaking type inference for Typescript modules dependent on these...

Issue #302, opened 2022-06-17
for `createPerson`, allow to associate the current users AgentPubKey optionally with the person being created


Connoropolous:
on the holochain side, do this by creating a link between the current users AgentPubKey and the newly created Agent entry, by its address.

weswalla:
can I assume there is some boolean flag passed in to the zome function, something like `link_pub_key: boolean`? @Connoropolous

Connoropolous:
yes that's right

Connoropolous:
for now, just trial it by hardcoding the underlying inner zome call, in the mutations code

pospi:
Would it be possible to also have it as an independent API endpoint? The flag can stay, or not- unfussed. That way newcomers can claim already-existing profiles and the like.

I suppose we also need some validation to prevent more than 1 profile being linked to the same `AgentPubKey`, but I'm guessing that's already done (just mentioning for completeness).

In addition to the validation, a useful behaviour might be to clear any existing association before setting up a new one when this "independent API endpoint" is called to make a profile association. That would be an easy way to facilitate folks updating their profiles with new data.

Connoropolous:
"I suppose we also need some validation to prevent more than 1 profile being linked to the same AgentPubKey, but I'm guessing that's already done (just mentioning for completeness)."
That validation isn't going to be possible, due to network partitions. We can't validate that kind of thing it's non-deterministic.

The most useful way to handle this is on the "read" side, not the "write" side I think. It might even be more accurate to make `myAgent` return an array. I'm not sure we can enforce singularity, or if we do, it's superficial. I guess we can enforce singularity within a single agents source chain. We just can't enforce singularity across the DHT.

Connoropolous:
@pospi if we do it with a separate API endpoint, how do we create a graphql layer for that, since its not in the Vf Graphql spec?

pospi:
Please see https://github.com/h-REA/hREA/issues/311#issuecomment-1164308123 in answer to the above :slightly_smiling_face:

Connoropolous:
closing this issue, as we are deciding to go with #311 instead

PR #307, opened 2022-06-22
associate my agent & get my agent

### LINKED ISSUE #298: Implement simple inverse queries of `provider` and `receiver` on agent records
A [PR to VF GraphQL](https://lab.allmende.io/valueflows/vf-schemas/vf-graphql/-/issues/100) has to be made first before any of the simple [inverse query edges](https://www.valueflo.ws/specification/inverses/) can be implemented.

You can ignore
- any fields involving `X.inScopeOf` for now, pending #297 
- any fields involving `EconomicResource.primaryAccountable` for now, pending #133 

This only includes "inverse queries", not "other queries".

The focus here is:
`Commitment.provider` and `Commitment.receiver`
`Intent.provider` and `Intent.receiver`
`EconomicEvent.provider` and `EconomicEvent.receiver`

- [ ] comment out [the lines so far in `Agent`](https://github.com/h-REA/hREA/blob/5f01158a78266d4c0e96b8ff34a3a0047f0a7b1f/modules/vf-graphql-holochain/resolvers/agent.ts#L46-L85) which will be uncommented later in #297 and #133
- [ ] add `Agent.commitmentsAsProvider`
- [ ] add `Agent.commitmentsAsReceiver`
- [ ] add `Agent.intentsAsProvider`
- [ ] add `Agent.intentsAsReceiver`
- [ ] add `Agent.economicEventsAsProvider`
- [ ] add `Agent.economicEventsAsReceiver`

### For `Commitment.provider`:
We will need these to be implemented:
https://github.com/h-REA/hREA/blob/5f01158a78266d4c0e96b8ff34a3a0047f0a7b1f/zomes/rea_commitment/zome_idx_planning/src/lib.rs#L35-L36

### For `Intent.provider`:
we already have
https://github.com/h-REA/hREA/blob/5f01158a78266d4c0e96b8ff34a3a0047f0a7b1f/zomes/rea_intent/zome_idx_planning/src/lib.rs#L30-L32
and
https://github.com/h-REA/hREA/blob/7e3d5ad7a0849f8c4996d9c767ce64c486058870/zomes/rea_intent/lib/src/lib.rs#L37-L42

is it then just a matter of hooking up the resolver function? Is this relationship tested?

### LINKED ISSUE #301: implement `type` field on EntryData of zomes/rea_agent
for this, we may not actually rely on the addition of `type` to `Agent` field upstream in vf-graphql. This will instead just be an 'internal implementation detail'. We may have to extend the native Agent typescript type, like

`type InternalAgentType = Agent & { type: string }` or something like that. For use in the zome call generic type signatures

### LINKED ISSUE #303: implement `getMyAgent` query, and backend zome fn
This depends on #302 

Do a lookup from the current users AgentPubKey via `get_links` to any related Agent record. Just take the first one and return it. We will naively assume one-key one-person for now

### LINKED ISSUE #311: Implement agent profile association for existing profiles
This is an replacing #302 needed for compatibility with the GraphQL spec.

After some discussion with @mayel and the Bonfire team we have decided that signup / authorisation flows are too technology-specific to become part of the core spec. Bonfire has [several custom API endpoints](https://gist.github.com/mayel/880aea55a7a1b3a295fe036235636f82) that deal with registration, authentication and authorisation; in their case specifically `createUser` transparently creates an REA `Agent` and other linked records.

hREA will need to do things differently, given there is no notion of "registration" in Holochain. It doesn't matter much how we deviate in these registration actions, so long as it all results in the `myAgent` query returning what has been associated with the active authentication. It is also true that any UI components built for signup & login will be technology-specific, so there is little point in standardisation in this area. GraphQL should ideally treat authorisation & authentication transparently in any case.

On this basis we should add this custom GraphQL mutations for hREA only: 

    associateMyAgent(agentId: ID!): Boolean!

...but the first one is ideal since it has the greatest flexibility to allow for #310 in future.

weswalla:
will need better case handling around link creation
closes #303 
closes #311 
closes #301 
closes #298 

- [x] updates (big breaking change)  to "@valueflows/vf-graphql": "0.9.0-alpha.6",

### DIFF
--- a/zomes/rea_agent/lib/src/lib.rs
+++ b/zomes/rea_agent/lib/src/lib.rs
@@ -13,8 +13,10 @@ use hdk_records::{
         create_record,
         read_record_entry,
         update_record,
-        delete_record,
-    },
+        delete_record, read_record_entry_by_header,
+    }, agent_info, links::create_link,
+    get_links, HdkLinkType, DataIntegrityError,
+    WasmError,
 };
 // use hdk_semantic_indexes_client_lib::*;
 
@@ -27,7 +29,27 @@ pub fn handle_create_agent<S>(entry_def_id: S, agent: CreateRequest) -> RecordAP
     where S: AsRef<str>
 {
     let (header_addr, base_address, entry_resp): (_,_, EntryData) = create_record(&entry_def_id, agent)?;
+    // TODO: check if contains `link_pub_key` field
+    let pub_key = agent_info()?.agent_latest_pubkey;
+    // TODO: error handling to match expect error type
+    let _pub_key_link = create_link(pub_key, header_addr.clone(), HdkLinkType::Any, ())?;
     construct_response(&base_address, header_addr, &entry_resp, /*get_link_fields(&base_address)?*/)
+}
+
+pub fn handle_get_my_agent() -> RecordAPIResult<ResponseData>
+{
+    let my_pub_key = agent_info()?.agent_latest_pubkey;
+    let mut links = get_links(my_pub_key, None)?;
+    let maybe_link = links.pop();
+    // assumes there is only one link
+    match maybe_link {
+        Some(link) => {
+            let header_hash: HeaderHash = link.target.into();
+            let (base_address, entry) = read_record_entry_by_header::<EntryData, EntryStorage, _>(&header_hash)?;
+            construct_response(&base_address, header_hash, &entry, /*get_link_fields(&base_address)?*/)
+        },
+        None => Err(DataIntegrityError::Wasm(WasmError::Guest(String::from("agent pub key link doesn't exist"))))
+    }
 }
 
 pub fn handle_get_agent<S>(entry_def_id: S, address: AgentAddress) -> RecordAPIResult<ResponseData>
@@ -47,6 +69,8 @@ pub fn handle_update_agent<S>(entry_def_id: S, agent: UpdateRequest) -> RecordAP
 
 pub fn handle_delete_agent(revision_id: HeaderHash) -> RecordAPIResult<bool> {
 
+    // load the record to ensure it is of the correct type
+    let (_base_address, _entry) = read_record_entry_by_header::<EntryData, EntryStorage, _>(&revision_id)?;
     // This is where indexes would be updated if necessary
 
     delete_record::<EntryStorage>(&revision_id)
--- a/zomes/rea_agent/lib/src/lib.rs
+++ b/zomes/rea_agent/lib/src/lib.rs
@@ -88,6 +88,8 @@ fn construct_response<'a>(
         // proposals,
         commitments_as_provider,
         commitments_as_receiver,
+        intents_as_provider,
+        intents_as_receiver,
     ): (
         // Vec<CommitmentAddress>,
         // Vec<EconomicEventAddress>,
@@ -98,6 +100,8 @@ fn construct_response<'a>(
         // Vec<ProposalAddress>,
         Vec<CommitmentAddress>,
         Vec<CommitmentAddress>,
+        Vec<IntentAddress>,
+        Vec<IntentAddress>,
     ),
 ) -> RecordAPIResult<ResponseData> {
     Ok(ResponseData {
@@ -118,6 +122,8 @@ fn construct_response<'a>(
             // proposals: proposals.to_owned(),
             commitments_as_provider: commitments_as_provider.to_owned(),
             commitments_as_receiver: commitments_as_receiver.to_owned(),
+            intents_as_provider: intents_as_provider.to_owned(),
+            intents_as_receiver: intents_as_receiver.to_owned(),
         }
     })
 }
@@ -140,6 +146,8 @@ fn get_link_fields(base_address: &AgentAddress) -> RecordAPIResult<(
     // Vec<ProposalAddress>,
     Vec<CommitmentAddress>,
     Vec<CommitmentAddress>,
+    Vec<IntentAddress>,
+    Vec<IntentAddress>,
 )> {
     Ok((
         // read_index!(agent(base_address).commitments)?,
@@ -151,5 +159,7 @@ fn get_link_fields(base_address: &AgentAddress) -> RecordAPIResult<(
         // read_index!(agent(base_address).proposals)?,
         read_index!(agent(base_address).commitments_as_provider)?,
         read_index!(agent(base_address).commitments_as_receiver)?,
+        read_index!(agent(base_address).intents_as_provider)?,
+        read_index!(agent(base_address).intents_as_receiver)?,
     ))
 }
--- a/zomes/rea_agent/lib/src/lib.rs
+++ b/zomes/rea_agent/lib/src/lib.rs
@@ -90,6 +90,8 @@ fn construct_response<'a>(
         commitments_as_receiver,
         intents_as_provider,
         intents_as_receiver,
+        economic_events_as_provider,
+        economic_events_as_receiver,
     ): (
         // Vec<CommitmentAddress>,
         // Vec<EconomicEventAddress>,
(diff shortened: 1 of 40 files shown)


zomes/rea_agent/lib/src/lib.rs
```diff
@@ -27,9 +29,29 @@ pub fn handle_create_agent<S>(entry_def_id: S, agent: CreateRequest) -> RecordAP
...
+    // TODO: check if contains `link_pub_key` field
+    let pub_key = agent_info()?.agent_latest_pubkey;
+    // TODO: error handling to match expect error type
+    let _pub_key_link = create_link(pub_key, header_addr.clone(), HdkLinkType::Any, ())?;
     construct_response(&base_address, header_addr, &entry_resp, /*get_link_fields(&base_address)?*/)
 }
 
+pub fn handle_get_my_agent() -> RecordAPIResult<ResponseData>
+{
+    let my_pub_key = agent_info()?.agent_latest_pubkey;
```
pospi, line 63: Neat! I didn't expect you to do all of this in the backend, but I'm glad it's happened that way.
Another option would have been to use the generic 'whois' method and pass our own `AgentPubKey` out of the cell and back in again at the resolver level. Either way I expect the code can be genericised to serve both needs.

Connoropolous: Can you explain (both to me and Wesley) what the `whois` function is? It is not in the VF Spec I don't think. What purpose does it serve?

Connoropolous: I am going to move this convo to an issue actually

Connoropolous: #309

weswalla: wouldn't we still need some backend solution to resolve an agent pub key to the agent record entry?

zomes/rea_agent/lib/src/lib.rs
```diff
@@ -47,6 +69,8 @@ pub fn handle_update_agent<S>(entry_def_id: S, agent: UpdateRequest) -> RecordAP
 
 pub fn handle_delete_agent(revision_id: HeaderHash) -> RecordAPIResult<bool> {
 
+    // load the record to ensure it is of the correct type
+    let (_base_address, _entry) = read_record_entry_by_header::<EntryData, EntryStorage, _>(&revision_id)?;
```
Connoropolous, line 94: does this follow a pattern from somewhere else? how'd this come up?

zomes/rea_agent/lib/src/lib.rs
```diff
@@ -27,9 +29,29 @@ pub fn handle_create_agent<S>(entry_def_id: S, agent: CreateRequest) -> RecordAP
...
     construct_response(&base_address, header_addr, &entry_resp, /*get_link_fields(&base_address)?*/)
 }
 
+pub fn handle_get_my_agent() -> RecordAPIResult<ResponseData>
+{
+    let my_pub_key = agent_info()?.agent_latest_pubkey;
+    let mut links = get_links(my_pub_key, None)?;
+    let maybe_link = links.pop();
+    // assumes there is only one link
+    match maybe_link {
```
Connoropolous:
```suggestion
    // assumes there is only one link
    match links.pop() {
```

zomes/rea_agent/lib/src/lib.rs
```diff
@@ -27,9 +29,29 @@ pub fn handle_create_agent<S>(entry_def_id: S, agent: CreateRequest) -> RecordAP
     where S: AsRef<str>
 {
     let (header_addr, base_address, entry_resp): (_,_, EntryData) = create_record(&entry_def_id, agent)?;
+    // TODO: check if contains `link_pub_key` field
+    let pub_key = agent_info()?.agent_latest_pubkey;
+    // TODO: error handling to match expect error type
```
Connoropolous: can this todo be done? anything blocking this? is it actually necessary?

zomes/rea_agent/lib/src/lib.rs
```diff
@@ -27,9 +29,29 @@ pub fn handle_create_agent<S>(entry_def_id: S, agent: CreateRequest) -> RecordAP
     where S: AsRef<str>
 {
     let (header_addr, base_address, entry_resp): (_,_, EntryData) = create_record(&entry_def_id, agent)?;
+    // TODO: check if contains `link_pub_key` field
```
Connoropolous: This one will be contingent on our decision on how to approach who/how/when the key and entry get linked

zomes/rea_agent/lib/src/lib.rs
```diff
@@ -27,9 +29,29 @@ pub fn handle_create_agent<S>(entry_def_id: S, agent: CreateRequest) -> RecordAP
...
+    let mut links = get_links(my_pub_key, None)?;
+    let maybe_link = links.pop();
+    // assumes there is only one link
+    match maybe_link {
+        Some(link) => {
+            let header_hash: HeaderHash = link.target.into();
+            let (base_address, entry) = read_record_entry_by_header::<EntryData, EntryStorage, _>(&header_hash)?;
+            construct_response(&base_address, header_hash, &entry, /*get_link_fields(&base_address)?*/)
+        },
+        None => Err(DataIntegrityError::Wasm(WasmError::Guest(String::from("agent pub key link doesn't exist"))))
```
Connoropolous: we could improve this error message just a bit. It should include info on how to improve the situation, but let's do that once we know the exact strategy for linking

Connoropolous (commented):
nice with the added tests, this is ripe for merge, as soon as we have our strategy and implement the 'linking' approach

zomes/rea_agent/lib/src/lib.rs
weswalla: yeah that is the pattern in other `handle_delete_{}` functions, I had copied it over when playing around and it looks like it will be needed once there are links off of agent entries to other entry types, and so decided to leave it.

zomes/rea_agent/lib/src/lib.rs
Connoropolous: so the decision was that there needs to be a new zome fn exposed endpoint. fn signature will be that it takes an AgentAddress and returns Ok(()) if succesful

zomes/rea_agent/lib/src/lib.rs
Connoropolous: @weswalla instead of this,  define a new error variant on `DataIntegrityError`. 

Use something like this one:
```
    #[error("No Agent data is associated with the currently authenticated user")]
    AgentNotLinked,
```

and then use `DataIntegrityError::AgentNotLinked` here instead.

zomes/rea_agent/lib/src/lib.rs
Connoropolous: So instead of affiliating this here, create a new zome fn

Connoropolous (changes requested):
There are two things we now need here in this PR, to accomplish #311 and #302. 

- adding the new zome function for `associate_my_agent` which follows the fn signature described in #311
- `associate_my_agent` should fail an attempt to associate an agent when one has already been associated. This will restrict each user to one and only one associated agent.

zomes/rea_agent/lib/src/lib.rs
pospi: Not for hREA, but we will eventually need it for non-REA parts of an application which deal with Holochain-native identifiers (read: `AgentPubKey`). So that's what `whois` will be for, and why it's not needed for MMR (:

zomes/rea_agent/lib/src/lib.rs
pospi: Good one. When we eventually have "DNA dependencies" in core there will be referential integrity checks against agent profiles stored in the dependent Agent cell and so those will end up throwing this error as well.

Connoropolous:
@pospi I removed the 'arghs' lol

PR #308, opened 2022-06-22
added agent_type field to rust side and in create agent mutations

weswalla:
- [x] does agent type need to be added to vf graphql definitions?

### DIFF
--- a/modules/vf-graphql-holochain/mutations/agent.ts
+++ b/modules/vf-graphql-holochain/mutations/agent.ts
@@ -45,7 +45,7 @@ export interface OrganizationUpdateArgs {
 export type updateOrganizationHandler = (root: any, args: OrganizationUpdateArgs) => Promise<OrganizationResponse>
 
 export interface AgentCreateArgs {
-    agent: OrganizationCreateParams & { agentType: String },
+    agent: OrganizationCreateParams & { agentType: string },
 }
 export interface AgentUpdateArgs {
     agent: OrganizationUpdateParams,
--- a/modules/vf-graphql-holochain/mutations/agent.ts
+++ b/modules/vf-graphql-holochain/mutations/agent.ts
@@ -45,7 +45,7 @@ export interface OrganizationUpdateArgs {
 export type updateOrganizationHandler = (root: any, args: OrganizationUpdateArgs) => Promise<OrganizationResponse>
 
 export interface AgentCreateArgs {
-    agent: OrganizationCreateParams,
+    agent: OrganizationCreateParams & { agentType: String },
 }
 export interface AgentUpdateArgs {
     agent: OrganizationUpdateParams,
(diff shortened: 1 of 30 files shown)


pospi:
> does agent type need to be added to vf graphql definitions?

No, not yet. We can continue using the `Person` & `Organization` APIs for now; just implementing "type" transparently in the resolver code so that both APIs are backed by the same Holochain zome.

For completeness, the related discussion in vf-graphql has started [here](https://lab.allmende.io/valueflows/vf-schemas/vf-graphql/-/issues/101) but there will be some coordination needed with Bonfire along the way, probably after we finish off these inverse query edge names and publish the next release (99% sure that will be tomorrow, sorry for the delay!)

pospi:
> use agent type for graphql error handling (e.g. query for person returning organization, etc)

I think this will be a matter of the zome logic switching on the value of that field, and possibly updating different indexes as a result. But that might only be the case once #268 is merged. If it doesn't come up, don't worry about it for now (:

modules/vf-graphql-holochain/mutations/agent.ts
```diff
@@ -45,7 +45,7 @@ export interface OrganizationUpdateArgs {
 export type updateOrganizationHandler = (root: any, args: OrganizationUpdateArgs) => Promise<OrganizationResponse>
 
 export interface AgentCreateArgs {
-    agent: OrganizationCreateParams,
+    agent: OrganizationCreateParams & { agentType: String },
```
Connoropolous:
```suggestion
    agent: OrganizationCreateParams & { agentType: string },
```

usually use lowercase `s` and first letters with typescript. I forget the exact nature of the difference, but I know to do this

zomes/rea_agent/rpc/src/lib.rs
```diff
@@ -26,6 +26,47 @@ pub use vf_attributes_hdk::{
...
+    {
+        let s = String::deserialize(deserializer)?;
+        match s.as_str() {
+            "Person" => Ok(AgentType::Person),
+            "Organization" => Ok(AgentType::Organization),
+            // TODO: error handling
+            _ => Ok(AgentType::Person),
+        }
+    }
+}
```
Connoropolous: I'm not sure about whether we want this to be an `enum`. I propose that we leave it as an open string for now, and don't worry about validating it.

zomes/rea_agent/storage/src/lib.rs
```diff
@@ -71,6 +73,7 @@ impl Updateable<UpdateRequest> for EntryData {
     fn update_with(&self, e: UpdateRequest) -> EntryData {
         EntryData {
             name: if !e.name.is_some() { self.name.to_owned() } else { e.name.to_owned().unwrap() },
+            agent_type: self.agent_type.to_owned(), //assume cannot update agent_type
```
Connoropolous, line 76: good assumption

Connoropolous (approved):
Looks good. On the `query` side of things, I suppose Graphql would just be totally unaware of this field and not return it on the response, as it would never be requested within the response portion of the graphql request. This field doesn't show it's utility yet, but will later

pospi:
> I suppose Graphql would just be totally unaware of this field and not return it on the response, as it would never be requested within the response portion of the graphql request. 

That's correct. So it will be an internal-use field- it will still come back in the response data within the resolvers probably, but once returned from those functions the GraphQL execution engine will prune it out.

PR #314, opened 2022-06-29
set up some nice default queries in hrea-graphql-explorer

Connoropolous:


pospi (commented):
I think it's a good starting point, but I might check to see if we can inject comments into the explorer as well as it will give folks a better idea. If not, I will merge.

pospi:
Put in some things but need to fix #315 in order to try them out.
If there are issues, we can revert 18d3dfd6

Connoropolous:
There should be a button to close the error overlay in the top right. I'm not saying thats a fix, but at least you can try what you added. And this PR didn't introduce the bug so we can also merge this and then separately I will tackle the bug, which will technically just mean talking with Jost and asking them for a fix I think.

apps/holorea-graphql-explorer/src/App.tsx
```diff
@@ -11,29 +11,112 @@ import bindSchema, { autoConnect, VfModule } from '@valueflows/vf-graphql-holoch
...
       hasPointInTime
     }
+    economicResource {
+      id
+      name
+      classifiedAs
+      accountingQuantity {
+        hasNumericalValue
+      }
+      onHandQuantity {
```
Connoropolous:
```suggestion
      onhandQuantity {
```

Connoropolous:
@weswalla what do you think of this? 
https://github.com/h-REA/hREA/pull/314/files#diff-22dcc3518890718f62f83a0abc86885d1fd9c0a94511afe5306c34b030586f8d

pospi:
have run through all this locally now, and works a charm

PR #336, opened 2022-07-12
Feature/220 ordered query apis

### LINKED ISSUE #331: Implement first pass at ordering of filtered resultsets in query APIs
A more complete set of requirements for this, given a page size of N:

- All `query_*` APIs must return records in a consistent sort order, no matter the query parameters provided
-   Different ordering will need to be done independently of the query parameters

Some work has already been done on this in https://github.com/h-REA/hREA/tree/feature/220-ordered-query-apis however it seems there are a few data architecture concerns to work through. Still figuring out the specifics, will split out additional issues if it turns out this is more involved than previously expected.

pospi:
MVP of ordering of filtered resultsets from search APIs.

All tests hopefully passing, but might be a few orderings to clear up still. Otherwise looks to be working well for use with baseline functionality.

Closes #331 but will also need to log a followup job to improve this later for alignment with original record creation times.


pospi:
So looks like some small errors to clear up in `conforming_resources.js` (which seems simple) and in `resource_logic.js` (which is mildly worrying but hoping it's just a mismatched ID)

lib/hdk_semantic_indexes/zome_derive/src/lib.rs
```diff
@@ -126,16 +126,21 @@ pub fn index_zome(attribs: TokenStream, input: TokenStream) -> TokenStream {
     let index_accessors = all_indexes.clone()
         .map(|(
             _index_type, _index_datatype, relationship_name,
-            _related_record_type_str_attribute,
+            related_record_type_str_attribute,
             related_index_field_type, related_index_name,
             _reciprocal_index_name,
         )| {
             let local_dna_read_method_name = format_ident!("_internal_read_{}_{}", record_type_str_attribute, relationship_name);
+            let remote_record_time_index_id: String = format!("{}.indexed", related_record_type_str_attribute);
```
Connoropolous: where does `.indexed` string originate from?

lib/hdk_semantic_indexes/zome_derive/src/lib.rs
```diff
@@ -154,6 +159,7 @@ pub fn index_zome(attribs: TokenStream, input: TokenStream) -> TokenStream {
                 "Remote" => format_ident!("index_{}_{}", record_type_str_attribute, relationship_name),
                 _ => panic!("expected index type of Local or Remote"),
             };
+            let remote_record_time_index_id: String = format!("{}.indexed", related_record_type_str_attribute);
```
Connoropolous: I think this is the question to my answer above. It'd be nice to move this into a helper function that generated this as it feels duplicate and would just be easy for it go getting misplaced in the code.

lib/hdk_time_indexing/src/index_tree.rs
```diff
@@ -144,14 +144,14 @@ pub (crate) fn get_index_segments(time: &DateTime<Utc>) -> Vec<IndexSegment> {
 /// Returns a `TimeIndexingError::Malformed` if an invalid link tag is passed.
 ///
 fn decode_link_tag_timestamp(tag: LinkTag) -> TimeIndexResult<DateTime<Utc>> {
-    let bits = tag.as_ref().split(|byte| { *byte == 0x0 as u8 });
+    let bits: Vec<&[u8]> = tag.as_ref().splitn(2, |byte| { *byte == 0x0 as u8 }).collect();
```
Connoropolous, line 148: I have a hard time reading the code on this line, as it uses some expressions and syntaxes I'm not familiar with. Could you put a code comment on it to say a snippet about what this does?

lib/hdk_time_indexing/src/sorting.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    Box::new(move |a, b| {
+        let a_timestamp = get_time_for_entry_hash(prefix.to_owned(), a);
+        let b_timestamp = get_time_for_entry_hash(prefix.to_owned(), b);
+        b_timestamp.cmp(&a_timestamp)
+    })
+}
+
+// :TODO: consider an appropriate way of handling errors
+// :TODO: should we adjust for possibility of entries being validly indexed multiple times in the same index?
```
Connoropolous: I'm a bit behind nowadays on how LinkTag works... can you add a little code comment with context for the index_link_prefix?

lib/hdk_time_indexing/src/sorting.rs
```diff
@@ -0,0 +1,62 @@
...
+ * `index_name` time-ordered index.
+ *
+ * It is presumed that the entries are validly linked in the index. If
+ * not, they will be placed at the end of the list in undefined order.
+ *
+ * Usage:
+ *  let mut my_entry_hashes: Vec<EntryHash> = some_list_of_hashes;
+ *  my_entry_hashes.sort_by(sort_entries_by_time_index("my_time_index_name"));
+ */
+pub fn sort_entries_by_time_index<'a, I>(index_name: &'a I) -> Box<dyn for<'r> Fn(&EntryHash, &EntryHash) -> Ordering + 'a>
```
Connoropolous, line 22: I'm feeling on the fence about this function. It occurs to me we could be doing the `get_links` call multiple times per entry_hash in the list of addresses. Could we, in the code that calls this function, instead call `get_time_for_entry_hash` a guaranteed once per address in the list, and then use that list of date structs paired with their entry hash to sort by?

Connoropolous (commented):
The approach overall seems great, I have a couple comment-y questions, and one core logic question 👍

lib/hdk_time_indexing/src/sorting.rs
pospi: We could, but I don't think it's necessary for now.

a) Rust does seem to do quite intelligent optimisations for core-level sorting APIs and the compiler might be optimising some of this for us; and
b) It does seem like (based on conversations on Discord with Paul D'Aoust) Holochain subconscious should optimise this for us and be fetching links efficiently from its local cache upon subsequent reads for the same data.

pospi:
Two remaining errors in the new plan tests you added recently @Connoropolous - care to investigate? Probably just ordering related or maybe a merge needs to be made.

Connoropolous:
@pospi I think a line that assigned a value to variable `processes` got removed here: https://github.com/h-REA/hREA/commit/90eb6a548b1962e700b0b24cefd0fafbc9ed8215#diff-222e219b24cfa0c22508edb248e1b19176f736760f7c485da1704494e823098bL158

and that is the source of the failing assertion

Connoropolous:
I've just pushed a fix for it, I think.

Connoropolous (approved):
Ah I missed that one. Sweet we are looking good to merge

PR #339, opened 2022-07-20
Feature/319 record uniqueness

### LINKED ISSUE #319: Inject random data into newly created records to ensure uniqueness.
Sometimes, different records have the same field values. We encountered this in #318 where separate `EconomicResources` might be created by associated `EconomicEvents` but end up with the same `EntryHash`, causing uniqueness problems that can affect indexing and other relationships between different bits of data.

To work around this, we should insert some random bytes into an internal storage field upon creating new records, and persist the same value between updates. These values should not be returned in any response data- they are purely for internal use.

This will apply to all record types except `ProposedIntent` and `ProposedTo`, which are idempotent entries in that they merely hold values to associate other record types.

pospi:
closes #319 

@Connoropolous feel free to reject this & suggest an appropriate algorithm if you think we need better randomness in the nonce- I did consider including some timestamp data in addition to random bytes but didn't want to get into the complexity of dealing with that unless we think it's warranted.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -733,6 +733,7 @@ version = "0.1.0"
 dependencies = [
  "holochain_serialized_bytes",
  "serde",
+ "serde_bytes",
  "serde_maybe_undefined",
  "vf_attributes_hdk",
  "vf_measurement",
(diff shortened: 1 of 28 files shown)


Connoropolous (approved):
This is good!

PR #340, opened 2022-07-21
#275 update to HDI and holochain 0.0.152

### LINKED ISSUE #275: After release of the first HDI (or integrity kit) of Holochain, update versions to that
here is an example to sample from: 
https://github.com/perspect3vism/agent-language/tree/main/hc-dna/zomes/tests

Connoropolous:
In case not using nix-shell, these are the package versions for this branch:
cargo install holochain --version 0.0.152
cargo install holochain_cli --version 0.0.50
cargo install lair_keystore --version 0.0.11

additionally you will need the following:
- Node v16
- pnpm
- install the `wasm32-unknown-unknown` target: `rustup target add wasm32-unknown-unknown`

Current HDK: 0.0.144

closes #275 

- [x] refactor bundles/dna_templates to have 'integrity' and 'coordinator' zomes
- [x] add updated dna-auth-resolver
- [x] set up nix-shell for holochain v0.0.150
- [x] update HDK versions to 0.0.142, instead of anywhere that it has 0.0.136 set, fix compilation issues
- [x] rename any reference from `Header` to `Action`
- [x] rename any reference from `Element` to `Record`
- [x] introduce entry enum type
- [x] separate out an integrity zome for each zomes
- [x] work those integrity zomes into the bundles/dna_templates
- [x] update @holochain/client (ask Connor about this)
- [x] define `IndexSegment` entry type and its method implementations in the integrity zome, or remove
- [x] define UUID entry type and wrapper enum in integrity, or remove (working with hashes only)
- [x] define separate link types in `hdk_records`, along with integrity zome crate (or define in link type in each object type integrity zome and pass in the type to hdk_records functions)
- [x] fix the integration tests


Connoropolous:
@pospi  FYI @weswalla is back and they're taking over this from me

Connoropolous:
we've just landed the migration on the Acorn project, so have learned a few lessons along the way and feel prepared for this

zomes/rea_agent/integrity_zome/src/lib.rs
```diff
@@ -0,0 +1,12 @@
...
+ *
+ * Defines the top-level zome configuration needed by Holochain's build system
+ * to bundle the app. This basically involves wiring up the helper methods from the
+ * related `_lib` module into a packaged zome WASM binary.
+ *
+ * @package Holo-REA
+ */
+
+pub use hc_zome_rea_agent_storage::EntryTypes;
+pub use integrity_types::LinkTypes;
```
weswalla: does this work? or do the procedural macros (`#[hdk_entry_defs]` etc) need to be invoked in this crate? @pospi @Connoropolous

weswalla: I suppose a different approach would be to define the `EntryTypes` enum in this file, and implement the `From<EntryStorage>` trait directly in here. I think this makes more sense actually.

Connoropolous:
@weswalla on this branch (just mirroring a change I made on `sprout` branch) delete the whole `_templates` and `example` folders? They're causing conflicts now, because I deleted them, but you changed them. Just delete them and we're fine.

weswalla:
Does this general integrity zome pattern look right?

bundles/dna_templates/agent/dna.yaml
```diff
@@ -26,11 +26,13 @@ integrity:
...
+      bundled: "<repository-path>/target/wasm32-unknown-unknown/release/hc_zome_rea_agent_integrity.wasm"
+
     # utility zomes
     - name: remote_auth_integrity
       bundled: "<repository-path>/target/wasm32-unknown-unknown/release/hc_zome_dna_auth_resolver_hrea_integrity.wasm"
-    - name: agent_integrity
-      bundled: "<repository-path>/target/wasm32-unknown-unknown/release/hc_zome_rea_agent_integrity.wasm"
 
 coordinator:
   zomes:
```
Connoropolous, line 36: You're going to need to add something like this, similar to how it is done for `remote_auth`
```
      dependencies:
        - name: agent_integrity
```

zomes/rea_agreement/integrity_zome/src/lib.rs
```diff
@@ -0,0 +1,12 @@
...
+ * Holo-REA agreement integrity zome for API definition
+ *
+ * Defines the top-level zome configuration needed by Holochain's build system
+ * to bundle the app. This basically involves wiring up the helper methods from the
+ * related `_lib` module into a packaged zome WASM binary.
+ *
+ * @package Holo-REA
+ */
+
+pub use hc_zome_rea_agent_storage::EntryTypes;
```
Connoropolous:
```suggestion
pub use hc_zome_rea_agreement_storage::EntryTypes;
```

bundles/dna_templates/agent/dna.yaml
```diff
@@ -1,34 +1,50 @@
...
+    # utility zomes
+    - name: remote_auth_integrity
+      bundled: "<repository-path>/target/wasm32-unknown-unknown/release/hc_zome_dna_auth_resolver_hrea_integrity.wasm"
+
+coordinator:
+  zomes:
+    # application zomes
+    - name: agent
+      bundled: "<repository-path>/target/wasm32-unknown-unknown/release/hc_zome_rea_agent.wasm"
+    - name: agent_index
```
Connoropolous, line 42: For all of this to work, there might need to be an 'index_integrity' zome as well, that gets added as a dependency to all the index zomes. It might be able to be the same zome for all of them

weswalla: are there any new integrity types that would need to get defined though? Can we not just import the `LinkTypes` from the same integrity zome as the `agent` zome does?

weswalla: I guess I was thinking we only need to define 1 integrity zome per dna, but if we are to split every zome into coordinator and integrity then we'd have 2

Connoropolous: Ah, I see. that could be true. Depends if the index zomes need entry types, or just link types. And if we were to make LinkTypes more specific then we'd need more specific integrity zomes. 

Your proposal could work

zomes/rea_agent/integrity_zome/src/lib.rs
Connoropolous: FYI @pospi 
Wesley and I jammed on this today and came up with some strategies

pospi:
Made a lot of progress, but a couple of new problems left to resolve:

- [x] `hc_zome_rea_unit_lib` isn't compiling and complains that its `LinkType`s aren't convertible to `ScopedZomeType<LinkType>`. I think this might be due to incorrect trait bounds in `anchored_record_helpers.rs`.
- [x] coordinator zomes which depend both on their own integrity zome crates *and* indirectly on `hc_zome_dna_auth_resolver_integrity` are causing linker errors about duplicate definitions of `entry_defs`. Two possible solutions, both requiring changes to the lib which you are welcome to make on its main branch:
    - create some intermediary crate defining the storage structs and declare them in each type-specific integrity zome in which they're needed (eg. `hc_zome_rea_resource_specification_integrity::EntryTypes`). In this case you will probably need to parameterise `hdk_records::rpc::call_zome_method` and its dependencies/dependants in order to specify the entry type struct.
    - **or** just make changes to `dna-auth-resolver` to use the built-in capability storage functionality of the HDK and do away with its integrity zome entirely.

weswalla:
the most recent commit (https://github.com/h-REA/hREA/pull/340/commits/e873e797ea545f9301b0226374ab46f9d4a76341) makes all crate compile, however, I think we will want to replicate the core/integrity split pattern of dna_auth_resolver (https://github.com/holochain-open-dev/dna-auth-resolver/blob/2c9951ae5cc567b7560773d94ec6c34a0afe0758/integrity/src/lib.rs#L9) because right now the `__num_entry_types` and `__num_link_types` callbacks are left out of the 2 crates listed below which I'm assuming will be an issue on the holochain side of things.

There are two places we will need to handle this:
1. `hc_zome_rea_agent_integrity` crate
2. `hc_zome_rea_economic_event_integrity` crate

With (1), the error occurs because we define a link type in the agent integrity zome, but the `hdk_semantic_indexes_integrity` crate already defines a `LinkTypes`. With (2), the error occurs because the `hc_zome_rea_economic_event_lib` crate depends on the `hc_zome_rea_economic_resource_lib` crate which leads to two separate entry type definitions using the `#[hdk_entry_defs]` macro.

One solution for (1) is to split the `hdk_semantic_indexes_integrity` crate into 2 crate, like in https://github.com/holochain-open-dev/dna-auth-resolver/blob/2c9951ae5cc567b7560773d94ec6c34a0afe0758/integrity/src/lib.rs#L9, however, this wouldn't address (2).

The other solution (that I can think of) is to do the split on the `zome` side rather than the `lib` side. Meaning both `hc_zome_rea_agent_integrity` and `hc_zome_rea_economic_event_integrity` are decomposed into 2 crates each (core/integrity), thus solving (1) and (2).

Connoropolous:
I did the split on both sides, semantic and classes and for the classes I did all of them instead of being selective

Connoropolous:
Next step is integration tests I think. Seeing how much we broke along the way

lib/hdk_indexing_core/src/lib.rs
```diff
@@ -0,0 +1,20 @@
...
+// pub fn __num_link_types() -> u8 {
+//     hdk_semantic_indexes_core::LinkTypes::len()
+// }
+
+#[hdk_link_types(skip_no_mangle = true)]
+pub enum LinkTypes {
+    IdentityAnchor,
+    SemanticIndex,
+    TimeIndex,
+}
```
Connoropolous: @pospi I'm moving your comment to the PR, as it's easier to track:

Pospi said
"hmm, I don't think there should be any situations where all 3 of these link types should be managed by the same controller zome. If this is just for Agent then it's possibly an architectural problem to resolve. I've been working steadily to separate those crates, ideally soon we can drop the semantic indexes dependency on hdk_records entirely and that will reduce this kind of confusion."

Connoropolous: My response...

But semantic and time indexing do get used in a colocated context right? So is it IdentityAnchor that would get isolated out?

Connoropolous:
Ok so, 

```
cd test
TRYORAMA_LOG_LEVEL=debug WASM_LOG=debug RUST_LOG="debug,wasmer_compiler_cranelift=error,holochain::core::workflow=error" RUST_BACKTRACE=1 npx tape agent/**/*.js
```

results in all failures, and a wasm related 'host/guest' interface issue:

```
error during scenario.addPlayerWithHappBundle:  {
  type: 'error',
  data: {
    type: 'internal_error',
    data: 'Conductor returned an error while using a ConductorApi: RibosomeError(WasmRuntimeError(RuntimeError { source: User(WasmError { file: "/Users/connor/.cargo/registry/src/github.com-1ecc6299db9ec823/holochain-0.0.152/src/core/ribosome/real_ribosome.rs", line: 416, error: Compile("Error while importing \\"__wbindgen_placeholder__\\".\\"__wbindgen_describe\\": unknown import. Expected Function(FunctionType { params: [I32], results: [] })") }), wasm_trace: [], native_trace:    0: <unknown>\n'
```

looking into the line throwing it:
https://github.com/holochain/holochain/blob/develop/crates/holochain/src/core/ribosome/real_ribosome.rs#L414

Connoropolous:
Oh! Damien posted about this issue recently and discovered the fix. Woohoo!
https://discord.com/channels/727959104777224303/728647693244694648/1005632262865432586
https://discord.com/channels/727959104777224303/728647693244694648/1005668887431487488

Connoropolous:
Pushed the fix, that is letting us get the integration tests running

weswalla:
It seems like the remaining errors (there may be others I haven't investigated yet) are related to `dna_auth_resolver`, as the following error is consistent across tests:
```
Err(Wasm(WasmError { file: "/Users/weswalla/.cargo/git/checkouts/dna-auth-resolver-bb2f8d37890db247/2c9951a/core/src/lib.rs", line: 32, error: Guest("AvailableCapability does not map to any ZomeId and LinkType that is in scope for this zome.") }))])
```

which is related to https://github.com/holochain-open-dev/dna-auth-resolver/blob/2c9951ae5cc567b7560773d94ec6c34a0afe0758/core/src/lib.rs#L32

where the error in generated in holochain: https://github.com/holochain/holochain/blob/29f7cf536c030c5d79c6595f4bdbcb0c61c3b9d1/crates/hdk_derive/src/link_types.rs#L71

lib/hdk_indexing_core/src/lib.rs
pospi: yep! thats right

Connoropolous: I'm going to leave it as a TODO I think

pospi: addressed this now, it's gone from the lib

Connoropolous:
merrrrrrrging

PR #344, opened 2022-07-26
remove out of date and empty _templates and examples

Connoropolous:
they just add extra top level folders to an otherwise already very full set of folders


Connoropolous:
I think I'm just going to do this, and if/when the time comes, git history can be used to reinstate any of these

PR #349, opened 2022-08-05
Feature/346 previous revisions metadata

### LINKED ISSUE #346: Return previous revision metadata with records to allow UIs to recurse backwards through record versions
`RecordMeta.previous_revision` needs to be returned in order for applications to have basic abilities to retrieve earlier versions and determine record provenance.

pospi:
Closes #346


Connoropolous (approved):
Looks solid. The presence (and passing) of the integration test is great 👍

PR #353, opened 2022-08-07
graphql-client-holochain API fixes and improvements

### LINKED ISSUE #350: fix input object type signature for vf-ui/graphql-client-holochain
Both 'conductorUri' and 'dnaConfig' should be optional, but aren't currently. Since that makes everything optional, the whole `options` object should be optional

This is important because this is the primary entry point for app devs

Connoropolous:
closes #350 by implementing a cleaner, bug free, and more well documented interface


pospi (approved):
looks good, and such better docs!
Although, seems to have broken the resolver generator function based on what the test output is saying- once that's fixed good to merge by me.

Connoropolous:
nice! one liner fix, pushed that.

PR #357, opened 2022-08-15
add onhand_effect to Action

### LINKED ISSUE #153: Add `onHandEffect` to `Action`
Connor attempt at work plan:

- add `onhand_effect` next to `resource_effect` here:
https://github.com/h-REA/hREA/blob/588b017242acbe284dde00eee0e468ed24c922e3/lib/vf_actions/src/builtins.rs#L25
- based on **[this reference](https://www.valueflo.ws/concepts/actions/#action-behaviors)**, add an extra `ActionEffect` parameter to each line item here: https://github.com/h-REA/hREA/blob/588b017242acbe284dde00eee0e468ed24c922e3/lib/vf_actions/src/builtins.rs#L38-L55
- adjust `get_event_action` so that it incorporates this new `onhand_effect` value of the builtin actions, and uses it instead if appropriate https://github.com/h-REA/hREA/blob/588b017242acbe284dde00eee0e468ed24c922e3/zomes/rea_economic_resource/storage/src/lib.rs#L294-L298

background info:
See https://github.com/valueflows/vf-graphql/pull/81

Connoropolous:
closes #153 

This PR is much easier to review in 'split diff' view instead of 'unified diff' view

tests are passing

### DIFF
--- a/lib/vf_actions/src/builtins.rs
+++ b/lib/vf_actions/src/builtins.rs
@@ -18,7 +18,7 @@ macro_rules! generate_builtin_actions {
                 stringify!($a) => Some(Action {
                     id: str::replace(stringify!($a), "_", "-"),
                     label: str::replace(stringify!($a), "_", "-"),
-                    resource_effect: $e,
+                    accounting_effect: $e,
                     onhand_effect: $f,
                     input_output: $g,
                     pairs_with: stringify!($h).to_string(),
@@ -29,9 +29,13 @@ macro_rules! generate_builtin_actions {
     }
 }
 
-// The first ActionEffect here is what is called the 'resource_effect' or
-// the 'accounting effect'. The second ActionEffect define the 'onhand_effect'.
-// They are separate because they can differ from each other for a given action.
+/*
+    The first ActionEffect here is what is called the 'accounting effect'.
+    The second ActionEffect is the 'onhand effect'.
+    They are distinct because some actions mark a resource as 'in use';
+    ie. it's not 'on hand' anymore, but it hasn't been removed from inventory
+    for accounting purposes and is still 'owned' by the same Agent(s).
+*/
 pub fn get_builtin_action(key: &str) -> Option<Action> {
     generate_builtin_actions!(
         key;
@@ -88,7 +92,7 @@ mod tests {
         let action = Action {
             id: "consume".to_string(),
             label: "consume".to_string(),
-            resource_effect: ActionEffect::Decrement,
+            accounting_effect: ActionEffect::Decrement,
             onhand_effect: ActionEffect::Decrement,
             input_output: ProcessType::Input,
             pairs_with: "notApplicable".to_string(),
--- a/lib/vf_actions/src/builtins.rs
+++ b/lib/vf_actions/src/builtins.rs
@@ -7,24 +7,21 @@
  *
  * @see https://github.com/valueflows/valueflows/issues/487#issuecomment-482161938
  */
-use super::{
-    Action,
-    ActionEffect,
-    ProcessType,
-};
+use super::{Action, ActionEffect, ProcessType};
 
 // setup for core actions as in-memory statics
 
 macro_rules! generate_builtin_actions {
-    ($key: expr; $( $a:ident => $e:expr, $f:expr, $g:expr );*) => {
+    ($key: expr; $( $a:ident => $e:expr, $f:expr, $g:expr, $h:expr );*) => {
         match &str::replace($key, "-", "_")[..] {
             $(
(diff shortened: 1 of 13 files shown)


lib/vf_actions/src/builtins.rs
```diff
@@ -7,52 +7,52 @@
...
                 })
             ),*,
             _ => None,
         }
     }
 }
 
+// The first ActionEffect here is what is called the 'resource_effect' or
+// the 'accounting effect'. The second ActionEffect define the 'onhand_effect'.
+// They are separate because they can differ from each other for a given action.
```
pospi: change to something like, "they are distinct because some actions mark a resource as 'in use'; ie. it's not 'on hand' anymore, but it hasn't been removed from inventory for accounting purposes and is still 'owned' by the same Agent(s)".
...Or ask @fosterlynn if she has a better explanation of the difference ;)

pospi (approved):
Looks great! Glad to have simplified this logic. Happy with whatever description of the attribute you & Lynn decide upon for the code comment :)

lib/vf_actions/src/builtins.rs
Connoropolous: if we want a deeper explanation maybe we can just provide that as a link to a reference instead of all this

fosterlynn: >...Or ask @fosterlynn if she has a better explanation of the difference ;)

That's basically right. :)  And I think better than what is there, because "resource effect" is confusing there.

>if we want a deeper explanation maybe we can just provide that as a link

I didn't find anything very succinct.  There is this page https://www.valueflo.ws/specification/event-resource/.  Or maybe the actions page https://www.valueflo.ws/concepts/actions/#action-behaviors.  Maybe I should write something in the latter page.

Connoropolous: "Maybe I should write something in the latter page."
I think there should definitely be something that equates the onhandEffect property of actions to the `onhandQuantity` of Resources. Of course it's obvious once you see it, and the wording conveys a connection, but it would be perfect for it to be explicit.

Connoropolous: ""resource effect" is confusing there."
I agree. Why DO we have it called resourceEffect there?
It obscures the relationship between itself and the "accountingQuantity"

fosterlynn: Right, resource effect can affect accounting and/or onhand quantities, depending on the action.  So it should be out of that statement.

Connoropolous: well, we don't currently have something in the code called "accounting effect", so should we replace "resource effect" with that? I'd like to if possible

lib/vf_actions/src/builtins.rs
```diff
@@ -7,52 +7,52 @@
...
 
 macro_rules! generate_builtin_actions {
-    ($key: expr; $( $a:ident => $e:expr, $f:expr, $g:expr );*) => {
+    ($key: expr; $( $a:ident => $e:expr, $f:expr, $g:expr, $h:expr );*) => {
         match &str::replace($key, "-", "_")[..] {
             $(
                 stringify!($a) => Some(Action {
                     id: str::replace(stringify!($a), "_", "-"),
                     label: str::replace(stringify!($a), "_", "-"),
                     resource_effect: $e,
```
Connoropolous: @fosterlynn this is where it's called "resource_effect"

fosterlynn: Not sure what that hunk of code actually does. :sweat_smile: .  Here is what I'm in process of putting in to the vf doc that you care about:

The economic resource's accounting quantity and onhand quantity are both affected (decremented or incremented or no effect) by the resource quantity of the economic event.

Hope that helps sort it out!

lib/vf_actions/src/builtins.rs
Connoropolous: my confusion was merited. it is called 
resourceEffect in vf-graphql as of now: https://lab.allmende.io/valueflows/vf-schemas/vf-graphql/-/blob/sprout/lib/schemas/action.gql#L26
I am renaming it internally to accounting_effect, but aliasing it to `resourceEffect` when an action gets passed to the client (graphql) so that it doesn't error. 
We should fix that in vf-graphql?

Connoropolous: I created an issue here: https://lab.allmende.io/valueflows/vf-schemas/vf-graphql/-/issues/103

fosterlynn: Yes, graphql and VF itself.  A breaking change, but ah well, this isn't the worst time for that.

PR #358, opened 2022-08-15
Feature/155 default unit of resource

### LINKED ISSUE #155: Add `ResourceSpecification.defaultUnitOfResource`
See https://github.com/valueflows/vf-graphql/pull/85

This is needed by rea-playspace project

Connoropolous:
fixes #155

### DIFF
--- a/modules/vf-graphql-holochain/resolvers/resourceSpecification.ts
+++ b/modules/vf-graphql-holochain/resolvers/resourceSpecification.ts
@@ -46,10 +46,11 @@ export default (enabledVFModules: VfModule[] = DEFAULT_VF_MODULES, dnaConfig: DN
       },
     } : {}),
     (hasMeasurement ? {
-      defaultUnitOfResource: () => {
-        // issue #155
-        // https://github.com/h-REA/hREA/issues/155
-        throw new Error('resolver unimplemented')
+      defaultUnitOfResource: async (record: { defaultUnitOfResource: AddressableIdentifier }) => {
+        if (!record.defaultUnitOfResource) {
+          return null
+        }
+        return (await readUnit({ id: record.defaultUnitOfResource })).unit
       },
       defaultUnitOfEffort: async (record: { defaultUnitOfEffort: AddressableIdentifier }): Promise<Maybe<Unit>> => {
         if (!record.defaultUnitOfEffort) {
--- a/zomes/rea_economic_resource/storage/src/lib.rs
+++ b/zomes/rea_economic_resource/storage/src/lib.rs
@@ -30,7 +30,7 @@ use vf_attributes_hdk::{
 };
 use vf_actions::{ ActionEffect, ActionInventoryEffect};
 pub use vf_actions::get_builtin_action;
-use hc_zome_rea_resource_specification_rpc::{ResponseData as ResourceSpecificationResponse};
+use hc_zome_rea_resource_specification_rpc::{ResponseData as ResourceSpecificationResponseData, Response as ResourceSpecificationResponse};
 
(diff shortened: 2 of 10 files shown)


modules/vf-graphql-holochain/resolvers/resourceSpecification.ts
```diff
@@ -46,10 +46,11 @@ export default (enabledVFModules: VfModule[] = DEFAULT_VF_MODULES, dnaConfig: DN
       },
     } : {}),
     (hasMeasurement ? {
-      defaultUnitOfResource: () => {
-        // issue #155
-        // https://github.com/h-REA/hREA/issues/155
-        throw new Error('resolver unimplemented')
+      defaultUnitOfResource: async (record: { defaultUnitOfResource: AddressableIdentifier }) => {
+        if (!record.defaultUnitOfResource) {
```
pospi, line 50: can use the destructured property here, no?

zomes/rea_economic_resource/storage/src/lib.rs
```diff
@@ -135,43 +135,70 @@ impl TryFrom<CreationPayload> for EntryData {
...
+                    None => default_unit_of_resource
+                };
+                Some(QuantityValue::new(resource_quantity.get_numerical_value(), unit))
+            },
+            _ => None,
+        };
+        // capture the unit_of_effort from any default
+        // set on the Resource Specification this conforms to,
+        // if any
+        let unit_of_effort = match maybe_resource_specification {
```
pospi: I think there's an error in this logic. `EconomicResource.UnitOfEffort` should come from the `EconomicEvent.effortQuantity.hasUnit` if set and fallback to the one from the associated `ResourceSpecification.defaultUnitOfEffort`, correct @fosterlynn?

That being the case I think we can also make this a bit more efficient by only querying the `ResourceSpecification` if the necessary fields aren't set in the `EconomicEvent`.

pospi (changes requested):

modules/vf-graphql-holochain/resolvers/resourceSpecification.ts
Connoropolous: would have to destructure the left side of this
`record: { defaultUnitOfResource: AddressableIdentifier }`

pospi: *shrugs* it's a minor thing :)

zomes/rea_economic_resource/storage/src/lib.rs
Connoropolous: sounds right. K

fosterlynn: > EconomicResource.UnitOfEffort should come from the EconomicEvent.effortQuantity.hasUnit if set and fallback to the one from the associated ResourceSpecification.defaultUnitOfEffort, correct @fosterlynn?

Well, it is hard to imagine an event that creates a resource having an effortQuantity.  So it should take the one from the resource specification as you suggested, if it is there, when created.  I also think we need the capability to set it directly as part of creating or updating a resource, which would take precedence over the one in the resource specification.  I also don't see a big problem with doing what you suggest with the effortQuantity.hasUnit if it is not already set.  (Not sure if we want to overwrite it in that situation, but could perhaps be persuaded. It seems like a user choice, and I'm actually not sure how many user groups would want it to be a firm setting, or just a default that can be overridden on particular events.)

Connoropolous: "So it should take the one from the resource specification as you suggested, if it is there, when created." this bit we already do

PR #359, opened 2022-08-15
fix stringIdRegex for matching unit identifiers so that it catches symbols and any non-digits

### LINKED ISSUE #354: stringIdRegex for Unit symbols is not loose enough for something like `$`
https://github.com/h-REA/hREA/blob/067ee13467a304140ca62deec084d7a4cfa7c6dc/modules/vf-graphql-holochain/connection.ts#L142

this regex doesn't pick up, and thus doesn't deserialize for zome calls, a Unit ID which is specified via an actual 'symbol' symbol, instead of a string. @pospi I am guessing this is not intentional

Connoropolous:
closes #354

### DIFF
--- a/modules/vf-graphql-holochain/connection.ts
+++ b/modules/vf-graphql-holochain/connection.ts
@@ -139,7 +139,11 @@ const HOLOHASH_PREFIX_AGENT = [0x84, 0x20, 0x24]   // uhCAk
 
 const serializedHashMatchRegex = /^[A-Za-z0-9_+\-/]{53}={0,2}$/
 const idMatchRegex = /^[A-Za-z0-9_+\-/]{53}={0,2}:[A-Za-z0-9_+\-/]{53}={0,2}$/
-const stringIdRegex = /^\w+?:[A-Za-z0-9_+\-/]{53}={0,2}$/
+// something like
+// $:uhC0k1mcUqQIbtT0mkdTldhBaAvR6KlKxIV2IYwJemHt-NO92uXG5
+// or kg:uhC0k1mcUqQIbtT0mkdTldhBaAvR6KlKxIV2IYwJemHt-NO92uXG5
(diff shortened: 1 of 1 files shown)


modules/vf-graphql-holochain/connection.ts
```diff
@@ -139,7 +139,11 @@ const HOLOHASH_PREFIX_AGENT = [0x84, 0x20, 0x24]   // uhCAk
 
 const serializedHashMatchRegex = /^[A-Za-z0-9_+\-/]{53}={0,2}$/
 const idMatchRegex = /^[A-Za-z0-9_+\-/]{53}={0,2}:[A-Za-z0-9_+\-/]{53}={0,2}$/
-const stringIdRegex = /^\w+?:[A-Za-z0-9_+\-/]{53}={0,2}$/
+// something like
+// $:uhC0k1mcUqQIbtT0mkdTldhBaAvR6KlKxIV2IYwJemHt-NO92uXG5
+// or kg:uhC0k1mcUqQIbtT0mkdTldhBaAvR6KlKxIV2IYwJemHt-NO92uXG5
+// but not 9:uhC0k1mcUqQIbtT0mkdTldhBaAvR6KlKxIV2IYwJemHt-NO92uXG5 (i.e. no digits in the id)
+const stringIdRegex = /^\D+?:[A-Za-z0-9_+\-/]{53}={0,2}$/
```
weswalla, line 146: I'm unsure what the `_+\-/` section is meant to capture, but since that hasn't changed in this commit I'm assuming its okay.

weswalla (approved):

modules/vf-graphql-holochain/connection.ts
pospi: it's just the base64 encoding that gets done, I believe those are valid characters.

pospi (approved):

PR #362, opened 2022-08-19
Feature/351 revisions api

### LINKED ISSUE #351: Add zome API endpoints for retrieving previous record revisions by `ActionHash`
API methods to retrieve specific (stale) revisions of records must be implemented for all relevant record types (basically exposing `read_record_entry_by_header` to the frontend). These should be linked to `revision(revisionId: ID!)` query edges for all record types (this is part of `VfModule.History`).

pospi:
closes #351

### DIFF
--- a/zomes/rea_agent/zome/src/lib.rs
+++ b/zomes/rea_agent/zome/src/lib.rs
@@ -47,6 +47,11 @@ fn get_agent(ByAddress { address }: ByAddress<AgentAddress>) -> ExternResult<Res
     Ok(handle_get_agent(address)?)
 }
 
+#[hdk_extern]
+fn get_revision(ByRevision { revision_id }: ByRevision) -> ExternResult<ResponseData> {
+    Ok(handle_get_revision(revision_id)?)
+}
+
 #[derive(Debug, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 struct WhoisParams {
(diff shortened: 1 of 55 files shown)


pospi:
Weird that this is timing out on some of the Ubuntu checks but not the MacOS one now. @Connoropolous any thoughts about that? Or do you think we should consider it working?

Connoropolous:
"Weird that this is timing out on some of the Ubuntu checks but not the MacOS one now. @Connoropolous any thoughts about that? Or do you think we should consider it working?"

I am not concerned about timeouts, so if any one of the 4 runs passes then I consider it a pass.

zomes/rea_agent/zome/src/lib.rs
```diff
@@ -47,6 +47,11 @@ fn get_agent(ByAddress { address }: ByAddress<AgentAddress>) -> ExternResult<Res
     Ok(handle_get_agent(address)?)
 }
 
+#[hdk_extern]
+fn get_revision(ByRevision { revision_id }: ByRevision) -> ExternResult<ResponseData> {
+    Ok(handle_get_revision(revision_id)?)
```
Connoropolous, line 52: I've been wondering if it makes any difference to write it this way, versus just: 
handle_get_revision(revision_id)

Connoropolous (approved):
All looks good. It's mostly just the same changes over and over again for each record type. I read over everything though 👍

zomes/rea_agent/zome/src/lib.rs
pospi: I could never get `Error`/`Result` conversion traits to work without the unwrap/rewrap. If you are able to, I'm interested in making it more succinct :wink:

Issue #264, opened 2022-04-25
Handle create_index! and update_index! calls properly, by not swallowing errors and giving sufficient error feedback


Connoropolous:
At the moment I have simply added debug calls so that we can at least see them fail, potentially, but this isn't sufficient. We need proper error handling.

https://github.com/h-REA/hREA/blob/3833a5c537321e5b06a6f6ad9b2f690c34991eee/zomes/rea_economic_event/lib/src/lib.rs#L190-L202

pospi:
Debug output is a nice first pass. It would be worth thinking about #176 whilst doing this; and might also be worth holding off until the new "integrity zomes" and "coordination zomes" distinction has landed in core (since it might provide integrity guarantees that we can leverage).

I'd also be curious to explore use of `post_commit` to manage index updates- this would ensure that no index failures could affect the related entry being written, which seems sensible (prevents malicious users creating malformed remote DNAs to cause a connected DNA to abort; though I also wonder how much damage such a bug could do since it would only affect the person configuring it). And also noting RE #176 the possibility that we'll want to handle validation of DNA-local and remote indexes differently.

The ideal would be that updating local index zomes can be considered to always succeed; and any possibilities of failure caught at compile & bundle time (see https://github.com/holochain/holochain/issues/976).

I suspect the only general-purpose solution for index zomes in foreign Cells is to treat them as non-critical and warn of errors. Unlike simpler Holochain apps, with hREA there *is* a distinction between "trusted" Cells configured as dependencies in the same collaboration space; vs "untrusted" Cells in neighbouring collaboration spaces. If it is someday possible to distinguish between the two at runtime (possibly via a hard dependency on a shared "cell registration" DNA) we might treat errors in the former as critical, and the latter as non-critical.

Once we can emit signals in `post_commit` this might be an appropriate way to alert of any indexing errors after writing a record. I think "alerting" is possibly all we can do, given the complex nature of intersecting multi-party networks.

pospi:
Made a pass at this for uniformity- all index updates and RPC calls are now logged, and treated as non-critical so as to avoid any cascading inter-cell request failures (which are kinda hard to debug). See 458daed9.

Connoropolous:
That's good progress. 

We could also change the response type of those functions, to include a section where those errors could be returned to the caller, and decisions made in that calling context about what, if anything, to do with that error

pospi:
[...]

PR #363, opened 2022-08-23
Feature/356 index integrity zome cleanup

### LINKED ISSUE #356: Rearchitect indexing integrity zomes to fix hidden link conflicts & storage bloat
With the changes to the HDK actioned in #340, data dependencies between controller zomes are now much more explicit than before and managed by regular Rust module system semantics. Where previously `get_links` used to return only links created in the local zome, now it can return links managed by other controller zomes if they are all acting upon the same integrity zome crates.

What this means for our use-case is that we ended up with duplicate links between entries in our semantic index zomes and had to move to an architecture of multiple distinct integrity zomes in order to keep the separation, per 1654968b5249cdc52e05153c37e6fdc092d94243.

However this creates (preserves?) a sub-optimal situation where locally-adjacent index zomes (eg. `hc_zome_rea_economic_event_index_observation` & `hc_zome_rea_process_index_observation`) are actually storing duplicates of the same data.

This is ideal & necessary for indexing that occurs in different cells, since *both* sides of an index need links to be made bidirectionally in order to run "read" & "query" actions&mdash; "reads" walk from a local record ID to the referenced ones and "queries" walk from referenced record IDs back to local ones. But in locally-adjacent indexes, these are duplicate sets of links.

It would be best for the sake of architectural simplicity if indexing code remained the same in both situations, as it makes composing indexes into various zome/cell combinations relatively straightforward.

So, the proposed path forward in order to optimise the storage efficiency of this situation whilst preserving the consistency of all semantic index type implementations is as follows:

- [x] revert the above commit, which will restore the system to a state where all indexing zomes share the same integrity zome crate.
- [x] alter the `create_index` behaviour in `hdk_semantic_indexes_zome_lib` to use `link_if_not_linked` instead of `create_link`

This should resolve all of the above, but needs to be tested. The end result will be such that for locally-adjacent indexes, an extra unnecessary API endpoint will be called that has no effect (since the other side of the index will have already created the links), but this is of minimal concern as it only affects the local node (no extra data will be replicated by the Holochain subconscious).

It's also worth noting that this foregoes a potentially useful future use-case for the semantic indexing library, whereby implementors might create multiple links between the same records in order to weight them. But it's probably fine to let this possibility go and not going to affect hREA.

pospi:
closes #356

It all turned out to be due to a single bug in the uniqueness of internal time index names used for remote records. They're not indexed by the relationship ID rather than the record ID.


Connoropolous:
sorry for the delay, I was traveling this week. I will give a review today!

Connoropolous:
the test failures here seem to be due not to timeouts but to some real issue with just a single test.

test/economic-resource/resource_validation.js
```diff
@@ -47,7 +47,7 @@ test('EconomicResource classification fields validation', async (t) => {
       },
     })
     await pause(100)
-
+    console.log('resp', JSON.stringify(resp))
```
Connoropolous:
```suggestion
```

test/economic-resource/resource_links.js
```diff
@@ -113,8 +113,8 @@ test('EconomicResource composition / containment functionality', async (t) => {
     readResource = readResp.economicResource
     t.deepLooseEqual(readResource.id, resourceId1, 'container resource re-retrieval OK')
     t.equal(readResource.contains && readResource.contains.length, 2, 'container resource reference appended')
-    t.deepLooseEqual(readResource.contains && readResource.contains[0], resourceId2, 'container resource reference B OK')
```
Connoropolous, line 116: The ordering of this seems to be the cause of the failing test :s

Connoropolous (changes requested):
we've just got this failing test on resource-links. I am not sure whether the test is wrong or the code.

test/economic-resource/resource_links.js
pospi: Pretty sure this is due to the following relationship, where reciprocals are declared in the same zome: https://github.com/h-REA/hREA/blob/sprout/zomes/rea_economic_resource/zome_idx_observation/src/lib.rs#L16-L17

I do seem to be observing similar issues with `EconomicEvent.affects` / `EconomicResource.affectedBy`, but they aren't showing up in tests (purely coincidentally). This also means we're still relying on duplicate indexes on the DHT in order to correctly query results.

I see two paths forward:

1. Delay this PR for now and revisit the scope to include index link validations outlined as part 1 of [this comment](https://github.com/h-REA/hREA/issues/264#issuecomment-1224314138).
2. Merge this as-is and log a bug for undefined ordering of locally-adjacent record indexes (eg. EconomicEvent <-> EconomicResource, Process <-> EconomicEvent, Proposal <-> ProposedIntent). Address the new bug with validation rules as per 1) above.

Either course of action would require migrations to the index integrity zomes format, but 2) results in more immediate gains of modest WASM bytesize savings.

Connoropolous: Hey sorry, I've been thinking about this. 
I think option `1` is better for this MMR release. The issue sounds interlinked with other ones and so a comprehensive sweep over them in a different work pass/scope sounds useful to get the job done proper.

Connoropolous: Is the fix in this commit useful to merge, on its own?
https://github.com/h-REA/hREA/pull/363/commits/f55783416fc0da9b108e5ca2c1599c0a541e069e

pospi: nah, doesn't matter. With the integrity zomes separated there can be no conflicts anyway, even though there will technically be conflicting/redundant links in different zomes within the same DHT.

pospi:
new commits! :crossed_fingers: if this passes tests, it's ready to merge.

pospi:
It's been a long road, but this one is now ready for merge! A semi-related smaller bug came up in the process, but more than happy to log that for later to be resolved with the larger indexing zome architecture cleanup to come (`post_commit` hooks et al).

lib/hdk_semantic_indexes/zome/src/lib.rs
```diff
@@ -344,13 +331,20 @@ fn create_dest_indexes<'a, A, B, S>(
...
 ) -> Box<dyn for<'r> Fn(&B) -> Vec<RecordAPIResult<ActionHash>> + 'a>
-    where S: 'a + AsRef<[u8]> + ?Sized + std::fmt::Debug,
+    where S: 'a + AsRef<[u8]> + ?Sized,
         A: DnaAddressable<EntryHash>,
         B: DnaAddressable<EntryHash>,
 {
     Box::new(move |dest| {
         match create_index(source, dest, link_tag, link_tag_reciprocal) {
-            Ok(created) => created,
+            Ok(created) => created.iter().cloned()
```
Connoropolous, line 342: If you can leave some developer facing code comments on this block of code then I'd be all sold on merging this!

Connoropolous (approved):
Awesome. Im glad that first commit was able to just be a revert. That worked out well. TBH the underlying issue here still going slightly over my head but entirely trusting your judgment on it.

lib/hdk_semantic_indexes/zome/src/lib.rs
Connoropolous: I just mean the whole filter map

Connoropolous:
Tests on Mac passing and the Linux tests failing but its on some upstream nix related issue. With Mac passing thats good enough.

lib/hdk_semantic_indexes/zome/src/lib.rs
pospi: I'll do you some better on code docs

pospi:
More on the underlying issue resolved in this PR, for clarity:

- Significant simplification of the integrity zome modularity & composability for indexing zomes. You can now use a single integrity zome definition per DNA which means you don't have to extend those Rust crates to create new indexes. It also means a much reduced possibility for future cell data migrations.
- Fixes for storage efficiency for semantic and time indexes- both are now much less likely to create redundant / duplicate DHT links, and tolerant of scenarios where it is unavoidable (eg. network partitions).
- Code reorganisation pass to keep all semantic indexing libs under the same subdirectory tree
- Some bonus docs and error types cleanup with thanks to @Connoropolous for prompting :slightly_smiling_face:

Connoropolous:
even better, awesome!

Connoropolous:
love this
Screen Shot 2022-10-19 at 12 27 28 PM

Connoropolous (commented):
Oh! Before merging this, I wonder if we can cancel out certain parts of the changes.

README.md
CHANGELOG.md
CONTRIBUTORS.txt

there is also a rename of the apps/holorea-graphql-explorer to hrea-graphql-explorer

apparently my original commit wasn't quite as clean as it should have been

Connoropolous:
closing as replaced by the same code but cleaned up #375

PR #365, opened 2022-09-02
HC bump 156

weswalla:
There are a whole bunch of trait bound error:

```
error[E0271]: type mismatch resolving `<EN as std::convert::TryFrom<AvailableCapability>>::Error == std::convert::Infallible`
   --> lib/hdk_semantic_indexes/client/src/lib.rs:476:8
    |
444 | fn request_sync_remote_index<EN, A, B, I, LT, E, E2>(
    |                                               - this type parameter
...
476 |     Ok(call_zome_method::<EN, _, _, _, _, _, _, _>(
    |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected enum `std::convert::Infallible`, found type parameter `E`
    |
    = note:        expected enum `std::convert::Infallible`
            found type parameter `E`
note: required by a bound in `call_zome_method`
   --> /Users/weswalla/sprillow/hREA/lib/hdk_records/src/rpc_helpers.rs:37:42
    |
37  |         EN: TryFrom<AvailableCapability, Error = E>,
    |                                          ^^^^^^^^^ required by this bound in `call_zome_method`
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -180,12 +180,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bfd4d1b31faaa3a89d7934dbded3111da0d2ef28e3ebccdb4f0179f5929d1ef1"
 dependencies = [
  "iana-time-zone",
- "js-sys",
  "num-integer",
  "num-traits",
  "serde",
  "time",
- "wasm-bindgen",
  "winapi",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -557,20 +557,9 @@ dependencies = [
 ]
 
 [[package]]
-name = "hc_zome_dna_auth_resolver"
-version = "0.1.0"
-source = "git+https://github.com/holochain-open-dev/dna-auth-resolver?rev=d0de3d0#d0de3d021f25427ef7214cbb666a529da604f114"
-dependencies = [
- "hc_zome_dna_auth_resolver_rpc",
- "hc_zome_dna_auth_resolver_storage",
- "hdk",
- "serde",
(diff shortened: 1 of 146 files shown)


Connoropolous (changes requested):
Actually @weswalla you had the right versions before the latest commit. Can we revert that commit? We will have to take a different approach to pinning the HDK dependency. Was it leaking in from dna-auth-resolver because of the HDK dependency being set to "0.0" there? I expect so.

weswalla:
@Connoropolous 
> Was it leaking in from dna-auth-resolver because of the HDK dependency being set to "0.0" there? I expect so.

Yes, exactly.

Connoropolous:
I will propose to Pospi that we specify the dep version over in dna-auth-resolver. This issue occurs regularly

weswalla:
@Connoropolous @pospi 
I forked dna-auth-resolver (https://github.com/weswalla/dna-auth-resolver) as a temp fix to see if it will compile, which is does.

Connoropolous:
for details on they 'why chrono' thing, see this commit: 
https://github.com/holochain/holochain/pull/1527/files

weirdly, you have to "disable" the wasm-bindgen feature by disabling all default features and then only enabling the ones that aren't bindgen. 
bindgen messes with the wasm in ways that breaks the holochain host/guest WASM interface

Connoropolous:
we can now run the integration tests

Connoropolous (approved):
bingo. 
luckily the tryorama update just got finished today. I didn't even realize till today that we'd be blocked on it.

PR #368, opened 2022-09-12
code organizing

Connoropolous:


pospi:
100% yes! This is cleanup that should've been done a while ago.

There are a couple of other things we can clean as well in this area.

- [ ] All statically defined `ByAddress` structs should be replaced with a reference to the templatized one exported from `hdk_semantic_indexes_zome_rpc` via `vf_attributes_hdk`.
- [ ] All use of the `ByAction` struct should be removed in favour of `ByRevision`, which makes more sense in the `hdk_records` context. 
    - [ ] Additionally, any delete or update API calls that are passing `address` as the parameter in the GraphQL layer will need to pass `revisionId` instead. But, I don't think there are any of these. It may just be lingering in the Rust layer in inter-cell RPC calls (eg. Satisfaction)

PR #369, opened 2022-09-13
Hc bump 161

### LINKED ISSUE #367: prior to Sapling release, update to holochain 0.0.161
I think it's just prudent to do this, due to the bugs they fixed. 

- [ ] will need to update dna-auth-resolver first

we should also update (forgot to do this) @holochain/client to v0.8.0 which included the network_seed change

weswalla:
closes #367

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,9 +1,9 @@
 let
-  holonixRev = "ce97dec539399d1e671ef7a2ee89e16de0b95561";
+  holonixRev = "0b1bf0eb19626f354c8efce70a21dbdbbaffbc71";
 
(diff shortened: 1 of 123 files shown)


Connoropolous (changes requested):
looking at the failing tests, I've seen those before.

Please try updating @holochain/client to 0.8.0 
and @holochain/tryorama to 0.8.0 and see if tests pass. 
I discovered this bug with tryorama, which could be the cause

https://github.com/holochain/tryorama/issues/145

Connoropolous (approved):
yes! tests passing

PR #375, opened 2022-10-21
Like PR 363 but the revert starts from latest `sprout` commit

### LINKED ISSUE #356: Rearchitect indexing integrity zomes to fix hidden link conflicts & storage bloat

Connoropolous:
original context in #363 
closes #356 

Pospi said: 
More on the underlying issue resolved in this PR, for clarity:

- Significant simplification of the integrity zome modularity & composability for indexing zomes. You can now use a single integrity zome definition per DNA which means you don't have to extend those Rust crates to create new indexes. It also means a much reduced possibility for future cell data migrations.
- Fixes for storage efficiency for semantic and time indexes- both are now much less likely to create redundant / duplicate DHT links, and tolerant of scenarios where it is unavoidable (eg. network partitions).
- Code reorganisation pass to keep all semantic indexing libs under the same subdirectory tree
- Some bonus docs and error types cleanup with thanks to @Connoropolous for prompting :slightly_smiling_face:


Connoropolous:
passing CI so let's go

PR #390, opened 2023-02-23
Feature/nix flakes update

pospi:
Not sure how this impacts CI, I guess the run commands will have changed?


pospi:
Ok so this is now a Holochain version update as well; FWIW I don't think `0.1.0` and `0.1.1` WASMs are compatible...

Connoropolous:
@pospi does this PR still matter? Do you want a review?

Issue #407, opened 2026-01-16
Add ResourceSpecification.mediumOfExchange


fosterlynn:
Just added this to Valueflows.  It is a boolean so you can tell resources that are the "real" stuff from the currencies, money, credits, tokens, etc. used for exchange on the reciprocal side.  Suggest default to false (not medium of exchange) if not entered.  https://www.valueflo.ws/specification/all_vf/#mediumOfExchange

PR #408, opened 2026-06-16
feat(vf-1.0): ValueFlows 1.0 + Holochain 0.7 (release branch for happ-0.5.0-beta)

### LINKED ISSUE #322: Query.offers & Query.requests
https://lab.allmende.io/valueflows/vf-schemas/vf-graphql/-/blob/sprout/lib/schemas/proposal.gql#L123-126

https://www.valueflo.ws/specification/inverses/#proposal

not intended as part of the MMR scope. Just capturing into the backlog for completeness for now

### LINKED ISSUE #398: Add field `substitutable` to ResourceSpecification
Since we removed the RecipeResource class, we need to add `substitutable` to ResourceSpecification.  It is a boolean, but also needs to be optional (like null or whatever holochain uses).  There isn't a logical default really.

It also then needs to be added to the graphql layer - both create and update.

Soushi888:
> **SoushAI analysis.** Drafted by Soushi's AI assistant, reviewed and posted by @Soushi888.

## What this is

The release branch for `happ-0.5.0-beta`: ValueFlows 1.0 on Holochain 0.7. The tag is cut from here, as `happ-0.4.0-beta` was cut from this same branch at `e35e5483` on 2026-07-27.

It began as an all-at-once staging branch. That phase is over: the work was split into a reviewable stack, each layer landed on its own, and the stack has now merged back in. Nothing is pending below this branch.

| PR | Carried | Merged |
|---|---|---|
| #409 | documentation, including a new consumer guide | 2026-09-10 |
| #410 | Tryorama retired, Sweettest at the zome boundary, regression suite | 2026-09-10 |
| #414 | zomes on `hdk 0.7.0` / `hdi 0.8.0`, holonix `main-0.7` | 2026-09-10 |
| #415 | `@holochain/client` 0.21, `hc-spin` 0.700, node 24, acceptance fixes | 2026-09-10 |

## Holochain 0.6.1 to 0.7

- `hdk 0.7.0` / `hdi 0.8.0` / `holochain_serialized_bytes 0.0.57`; holonix `main-0.7`, pinned in `flake.lock` at `ffcc7c63`
- every `validate_*` signature moves from `EntryCreationAction` to `TypedAction<EntryCreationData>` and its siblings
- `@holochain/client` 0.20 to 0.21 in the adapter and the UI, `hc-spin` 0.601.3 to 0.700.0, dev shell node 22 to 24
- `RUSTFLAGS='--cfg getrandom_backend="custom"'` is now mandatory for the wasm build: `hdi 0.8` pulls in `getrandom 0.3`, which refuses `wasm32-unknown-unknown` without a backend

**Tryorama is retired, and that is what unblocked 0.7.** The last published `@holochain/tryorama` is 0.19.2 and depends on `@holochain/client ^0.20.4`. No version targets 0.7 or client 0.21, so the old suite could not have run against this release under any arrangement.

**One 0.7 adapter bug found and fixed on the way.** 0.7 splits an Action into `{ header, data }`, moving `author`, `timestamp`, `action_seq` and `prev_action` under `header`. Anything reading `signed_action.hashed.content.timestamp` silently gets `undefined`, so `meta.retrievedRevision.time` was null on every record until `actionTimestamp()` learned the new path.

## VF 1.0 surface

New fields: `Proposal.purpose` (`offer` | `request`) with the `offers` / `requests` queries (closes #322), `ResourceSpecification.substitutable` (#398) and `mediumOfExchange` (#407), `EconomicEvent.effortQuantity` / `settles` / `reciprocalRealizationOf`, `Commitment.reciprocalClauseOf`.

New entry types with integrity, coordinator and GraphQL: `Claim` (with the `Claim.settledBy` reverse of `EconomicEvent.settles`), `SpatialThing`, `AgreementBundle`. New `EntryTypes` and `LinkTypes` variants are **appended**, never inserted, to preserve DHT index ordering.

**Corrections applied per @fosterlynn's review.** `ProductBatch` removed end to end (dropped from VF; `BatchLotRecord` replaces the concept). `Commitment.inScopeOf` resolver reverted, since VF 1.0 keeps `inScopeOf` on `Process` only. No `Appreciation` / `Settlement` / `Scenario`.

Worth stating plainly because it has been misread before: VF 1.0 **removed** `ProposedIntent`, `Satisfaction` and `Fulfillment` as separate classes, and its `Agreement` has no parties. hREA not having them is compliance, not a gap. The relations live on as direct fields: `EconomicEvent.fulfills` / `.satisfies`, with `Commitment.fulfilledBy` and `Intent.satisfiedBy` as reverse edges.

## Integrity validation

The integrity zome previously validated only `EconomicEvent` and `Proposal`; every other type was a permissive stub. This branch enforces the rules DHT-wide: temporal consistency, quantity non-negativity, required strings, the VF builtin action vocabulary, field immutability on update, WGS84 bounds on `SpatialThing`, `Intent.minimumQuantity <= availableQuantity`, bounded collections at `MAX_COLLECTION_LEN` 1024, and transfer-class actions requiring both `provider` and `receiver`.

Consequence for anyone upgrading: writes a `happ-0.4.0-beta` conductor accepted may now be rejected, with a message naming the rule.

## Tests

Two surfaces replace the retired Tryorama suite, at different layers. Both were run on the merged content.

**Sweettest, at the zome boundary:** 47 tests across six binaries, 0 failed, against a live 0.7 conductor. Every shared ValueFlows validator has a rejection test asserting on that rule's own message, and every rule with a boundary has an accepting case sitting on the boundary. Three of the tests were proven able to fail by mutating them, then reverted.

**Acceptance battery, at the GraphQL adapter:** `71/71 steps green` against a live 0.7 conductor on `@holochain/client` 0.21.

The 202 checks the Tryorama suite carried were mapped one by one against the replacements rather than compared by count. Nine behaviours had no assertion anywhere and are closed in `clients/acceptance/src/scenarios/regressions.ts`. A later module, `forward-refs.ts`, reads forward references back as objects, which the suite had never done: every reverse link was tested and almost no forward one.

## Bugs the new coverage found

- `EconomicResource.containedIn` was permanently null: the resolver read `record.contained_in`, a key `snakeToCamel` had already renamed, and the schema declares it singular while the resolver returned a list.
- `organization(id)` returned a Person and `person(id)` returned an Organization. Two bugs cancelling: the guards returned on mismatch instead of throwing, and compared `agentType` against lowercase strings the zome never writes.

Both were confirmed red against the unfixed adapter before the fix and green after.

## Test infrastructure

**The acceptance conductors are network-isolated now.** `hc sandbox generate network` exists on 0.7 and takes `mem` or `quic` plus `-b`, but it sets the transport and leaves peer discovery on `https://dev-test-bootstrap2.holochain.org/`. A validation dependency fetched with `must_get_valid_record` then goes out over that network and gives up at `request_timeout_s`, which is 60. That failed 21 of 65 CI steps with timeouts spaced exactly 60 seconds apart, while the same suite passed locally in 20 seconds. The battery now starts its own `kitsune2-bootstrap-srv` per run, which is the isolation Sweettest always had via `SweetConductor::standard()`.

**The acceptance runner used to hang after passing.** `hc sandbox --run` execs `holochain` as a grandchild, so killing the sandbox left the conductor holding an attached pipe and Node waited on it forever. CI showed `65/65` and was killed 86 minutes later. The runner now exits on its result, and the conductor is spawned detached so teardown reaps the whole process group.

**The cargo workspaces are unified.** `tests/sweettest` was its own workspace with its own `Cargo.lock`, which is how its conductor came to sit at 0.6.1 while the zomes moved to 0.7: two files recorded two versions and the upgrade read one. It is now a member that is not a `default-member`, so `cargo build --target wasm32-unknown-unknown` still builds only the zomes and the wasm graph contains no conductor crate.

**`release.yml` is fixed.** It called `yarn run build:holochain:release`, a script in no `package.json` on any branch, and published seven per-module DNA bundles from the retired multi-DNA layout. That is why the `happ-0.4.0-beta` run failed and its artifacts were uploaded by hand. It now builds with `build:happ`, uploads `workdir/hrea.happ` and `dnas/hrea/workdir/hrea.dna` behind `test -f` guards, and waits for `create-release` instead of racing it.

## Known limitations

- **`update_rea_economic_event` applies only `note`** and silently discards its other ten declared fields, returning success. It also makes three immutability rules unreachable from any public path. Pre-existing, not a regression. Tracked in #416.
- **`deleteEconomicEvent` does not exist**, by design: events are append-only, corrected with a compensating event. `EconomicResource` has no create or delete either; resources come into being through an event.
- `ResourceSpecification.substitutable` is stored by the zome and not yet exposed in the schema.
- `agentRelationship` and its role types are declared by the base ValueFlows schema and not implemented by this adapter. `clients/acceptance` carries a tripwire step that will fail the day they are.
- `QuantityValue.has_unit` referential validation is deferred: it adds a DHT read per quantity and deserves its own change.

## Follow-ups, not in this branch

- `BatchLotRecord`, replacing the removed `ProductBatch`.
- VF relational wiring: `EconomicResource.lot` to `BatchLotRecord` and `.currentLocation` to `SpatialThing`, currently opaque strings.
- Pending VF 1.0 deprecations, pre-existing and breaking, to discuss with @fosterlynn: removing `inScopeOf` from `EconomicEvent` / `Intent` / `Proposal` / `Plan`, and `atLocation` from `EconomicEvent` / `Commitment` / `Intent`.
- Legacy `pass` / `fail` actions, kept until their removal is agreed.

## Documentation

`docs/` is written against this merged state: [consuming a release](docs/consuming-a-release.md) for integrators, plus getting started, repository structure, architecture, GraphQL API and contributing. The consumer guide is the one to read before pinning: it covers the two artifacts, composing `hrea.dna` into your own `happ.yaml`, the version pinning contract, the 0.6 to 0.7 breaking changes, and what does and does not carry over from `happ-0.4.0-beta`.


Soushi888:
## Roadmap: remaining VF 1.0 work (revised per @fosterlynn's spec clarifications)

> **Updated.** Lynn reviewed the original plan and corrected several items against the actual VF 1.0 spec. This version folds those in. Net effect: some previously-planned work targets concepts that were **removed from / never in** VF 1.0, so it's been dropped or moved behind a discussion.

### Landed so far (full tryorama suite green: 18 files / 19 tests)
- Collection endpoints + connectivity: `get_all_commitments` / `get_all_intents` / `get_all_recipe_flows`, `Intent.satisfiedBy`, `agent(id)` fix.
- Integrity validation across 10 entry types (was 2): temporal consistency, quantity positivity, required fields, VF action validity, update immutability, range checks.
- `Claim.settledBy` reverse relation (kept as the 1.0 shape; a reified `Settlement` entity is **not** in 1.0).

### Corrections applied (from Lynn's review)
- **`ProductBatch` removed.** It was dropped from VF a while ago; `BatchLotRecord` replaces the concept (and `expiry_date` belonged to the old structure). The `ProductBatch` entry type added on this branch is being **removed**. `BatchLotRecord` (`batchLotCode`, `expirationDate`, `EconomicResource.ofBatchLot`) can be added later as its own focused PR.
- **`Commitment.inScopeOf` removed.** In VF 1.0 `inScopeOf` exists **only on `Process`**. The `Commitment.inScopeOf` resolver added on this branch is being reverted.

### Needs discussion with Lynn before any change (pre-existing, breaking)
These are pre-existing hREA fields that 1.0 removed, but consumers (R&O, Nondominium) may depend on them, so they are **not** being touched unilaterally:
- **`inScopeOf` on `EconomicEvent` / `Intent` / `Proposal` / `Plan`** — removed in 1.0 (only `Process` keeps it). Lynn: "if anyone is using the old ones, let's discuss."
- **`atLocation`** on `EconomicEvent` / `Commitment` / `Intent` — removed in 1.0.
- **Relational wiring (was "Wave 2"):** `EconomicResource.lot` → `BatchLotRecord` and `EconomicResource.currentLocation` → `SpatialThing` (currently opaque `String`s) remain valid 1.0 targets. The `atLocation` half of the original plan is dropped per the above.

### Not in VF 1.0 — dropped from the roadmap (was "Wave 3")
Per Lynn, these are **not** part of VF 1.0:
- **Appreciation** — not in 1.0. Dropped.
- **Settlement** — not in 1.0 (the `Claim.settledBy` link already covers the 1.0 settlement shape). Dropped.
- **Scenario** — not in 1.0 yet; it returns when the pre-operational / analytical part of the model is tested in a couple of projects. **Deferred** until VF reintroduces it; Lynn to advise on the shape.

### Deferred (deliberate, unchanged)
- `vf_actions` crate is hdk-dependent, so the integrity zome mirrors the builtin action ids as a const (`VF_BUILTIN_ACTIONS`) instead of importing it. Keep in sync.
- `QuantityValue.has_unit` referential validation — would add a DHT read per quantity to validation; worth its own focused change. Lynn is also considering an additional unit taxonomy alongside OM2, so this stays deferred.

@LeosPrograms @fosterlynn — flagging the "needs discussion" block especially. Happy to take Lynn up on the walkthrough offer before we cut the targeted PRs.

fosterlynn:
Clarifications, just in case (looking only at the above, not the actual changes).

>ProductBatch / BatchLotRecord

ProductBatch was removed quite awhile ago; BatchLotRecord replaces that concept.  You may understand this, but I also see a reference to expiry_date, which was part of the old data structure.

>Deferred: QuantityValue.has_unit referential validation (adds a DHT read per
quantity to validation — worth its own focused change).

Good to defer that, I'm considering adding another taxonomy for units, in addition to OM2.  Thinking about how this would work and be simple enough for devs/users.  Agree this is worth its own focused change.

>added the Commitment.inScopeOf resolver (it was the only scoped entity missing one)

All but one of the inScopeOf relationships were removed in 1.0, only Process remains.  But if anyone is using the old ones, let's discuss.

>atLocation / currentLocation → SpatialThing

In 1.0, atLocation was removed.

>Wave 3 — new VF 1.0 entry types (not started)

Appreciation, Settlement, Scenario are not in VF 1.0.  Scenario will be back when that part of the model (pre-operational, analytical) is tested in a couple projects.  I can update you on what we think it will be if needed.

@Soushi888 @LeosPrograms Happy to walk through anything you like.  Also, I'll check in discord for possible problems with 1.0 removals outside of what you two are working on.

Soushi888 (commented):
## Hermes Agent Review — COMMENT

Reviewed the full diff against `main-0.6`: 70 files, +4829/-317. All integrity zomes, coordinator zomes, the GraphQL adapter, the acceptance + e2e test surface, and the CI/release workflows. Both CI checks are green.

This is high-quality, coherent work. The integrity hardening — shared, reusable validators (temporal, quantity, action vocabulary, immutability, collection bounds, transfer-agent shape) wired consistently across create **and** update for every entry type — is exactly the DHT-wide validation this framework was missing. The append-only `EntryTypes`/`LinkTypes` variants (never inserted) correctly preserve existing DHT index ordering. `Proposal.purpose` immutability is enforced at the integrity layer with the right rationale documented (purpose-index link cleanup is deliberately skipped). The acceptance battery is genuinely end-to-end — real Apollo GraphQL against a live conductor, asserting accounting arithmetic, process I/O, commitment→fulfillment, intent→satisfaction, the full Claim→settle→settledBy chain, the offers/requests purpose partition, and negative paths for every integrity gate.

Two findings below — neither blocks this as a staging branch.

### Findings

**1. Dead code — `dnas/hrea/zomes/coordinator/hrea/src/rea_economic_event.rs` lines 340–~540**
An entirely commented-out duplicate of the old single-arg `create_rea_economic_event`. ~464 of the file's 1086 lines are commented (~43%). Adds noise and drift risk. Recommend deleting in the focused EconomicEvent follow-up PR — it's preserved in git history if ever needed.

**2. Referential-validation asymmetry — `dnas/hrea/zomes/integrity/hrea/src/rea_economic_event.rs:82`**
`validate_create_rea_economic_event` does DHT referential checks for `provider`, `realization_of`, `input_of`, `triggered_by` — but not for the new VF 1.0 relations `settles`, `reciprocal_realization_of`, `receiver`, `resource_inventoried_as`, `to_resource_inventoried_as`. The field-level validators (action, temporal, quantity, transfer-agents) all run, so this is a consistency gap, not a correctness hole. Worth a note in the known-follow-ups list so it survives the carve into focused PRs.

---
*Reviewed by Hermes Agent (Soushi888)*

Soushi888:
## Adapter publish to accompany `happ-0.4.0-beta` — status + proposal

Picking this up tomorrow. Leaving the state here so I (or anyone) can resume cleanly.

### The gap

`happ-0.4.0-beta` was tagged yesterday off this branch (`e35e5483`, which is on `feat/vf-proposal-purpose`). That DNA ships the full VF 1.0 purpose surface — `get_proposals_by_purpose` zome fn (`collections.rs:99`), `ReaProposal.purpose` with validation (`integrity/rea_proposal.rs:21`).

**But the npm adapter was not republished alongside it.** The newest version on npm is still `0.600.0-rc.0` (Dec 2025), which predates this PR's `vf1ProposalPurposeExtension`. So any consumer who installs `happ-0.4.0-beta` and then `bun install`s the published adapter gets a DNA that can answer `get_proposals_by_purpose` behind a client that doesn't know how to ask. The release pair is incoherent.

This bit me concretely in R&O: PR happenings-community/requests-and-offers#179 integrates `happ-0.4.0-beta`, confirms the reads work end-to-end, and correctly defers `Proposal.purpose` as "gated on the npm adapter shipping the extension" — because the published `0.600.0-rc.0` has no purpose resolvers, even though the DNA it pairs with does.

### What's already done (in the worktree, uncommitted)

- `modules/vf-graphql-holochain/package.json`: bumped `0.600.0-rc.0` → **`0.600.0-rc.1`**.
- `npm run build:graphql:adapter`: clean. `build/package.json` carries `0.600.0-rc.1`.
- Verified the artifact carries the extension: `build/index.js` has 7 refs to `vf1ProposalPurposeExtension` / `ProposalPurpose`; `build/src/queries/index.js` routes `offers`/`requests` → `get_proposals_by_purpose`.
- `npm publish --dry-run --tag rc`: clean. Tarball is `@valueflows/vf-graphql-holochain@0.600.0-rc.1`, 46.3 kB packed / 227.1 kB unpacked, 27 files.

### Two things npm caught that shape the real publish

1. **The `--tag rc` flag is mandatory** for a prerelease version. Using `rc` (not `latest`) keeps `0.600.0-rc.1` off the `latest` channel — existing `^0.600.0` consumers won't auto-upgrade to a prerelease. Consumers opt in by bumping explicitly. That's the right behaviour for a beta-accompanying release.
2. **The publish command** is:
   ```bash
   npm run build:graphql:adapter
   cd modules/vf-graphql-holochain/build
   npm publish --access=public --tag rc
   ```
   (`release.yml` doesn't publish the adapter automatically — it only ships the DNA on `happ-*` tag push. The adapter publish is the manual `publish:graphql:adapter` root script.)

### The open decision (why I stopped short of publishing)

This branch is **still a draft PR targeting `main-0.6`** — not merged. Publishing `0.600.0-rc.1` from an unmerged feature branch is workable (the tag points at a specific commit we control, the code is tested green here) but it means the published artifact diverges from the default branch. Two clean options:

- **Publish from the tag as-is** (pragmatic): the DNA is already out from `e35e5483`, so the adapter should match it semantically. Ship `0.600.0-rc.1` now, merge this PR to `main-0.6` on its own timeline. R&O and any other consumer gets purpose today.
- **Merge to `main-0.6` first, then publish** (tidier): one coherent target branch, tag and publish from there. More ceremony, but `main-0.6` won't be behind the published artifact.

I lean toward **publish from the tag**: the DNA already commits us to that commit's semantics, and an `-rc` tag honestly signals "not yet merged to main." Reconcile `main-0.6` separately.

### Resume plan

- [ ] Decide: publish-now-from-tag vs merge-then-publish.
- [ ] `npm adduser` on the publishing machine (auth isn't configured where I'm working).
- [ ] Run the publish command above.
- [ ] Verify: `npm view @valueflows/vf-graphql-holochain@0.600.0-rc.1 version`.
- [ ] Bump R&O's pin `^0.600.0-dev.0` → `0.600.0-rc.1`, add a purpose e2e assertion, close the gap declared in requests-and-offers#179.

The version bump sits in `.worktrees/feat-vf-proposal-purpose` as an uncommitted working-tree change, ready to commit whenever we publish.

Soushi888:
## Plan update: Holochain 0.7 and Sweettest, delivered as a stack on top of this PR

Both nondominium and requests-and-offers consume hREA, and nondominium's own 0.7 upgrade is currently waiting on an hREA build for 0.7. Rather than open a separate 0.7 branch and rebase this one onto it, the 0.7 upgrade will build on this branch, delivered as a stack of small dependent PRs rather than more commits piled onto this one.

### Why build on this branch rather than rebase onto a separate 0.7 branch

I measured the alternative before deciding. A three-way merge of this branch against a 0.7 port cut from `main-0.6` produces 12 conflicted files, 10 of them integrity zome modules. That is expected: this branch rewrote 15 of the 20 files in `dnas/hrea/zomes/integrity/hrea/src` for VF 1.0 compliance.

Meanwhile the 0.7 Rust change is almost entirely mechanical. In a prototype port, 426 of 426 changed lines across the `rea_*.rs` modules are signature substitutions:

```
EntryCreationAction  ->  TypedAction<EntryCreationData>
Update               ->  TypedAction<UpdateData>
Delete               ->  TypedAction<DeleteData>
CreateLink           ->  TypedAction<CreateLinkData>
DeleteLink           ->  TypedAction<DeleteLinkData>
```

No validation bodies change, because those parameters are unused in our validators. Applying that substitution once on top of the VF 1.0 work is strictly cheaper than resolving the same substitution across 12 conflicted files.

### The stack

This PR stays as it is. The remaining work lands as three dependent PRs above it, so each layer is reviewable on its own and nothing here grows further:

```
main-0.6
└── feat/vf-proposal-purpose            this PR, unchanged
    └── test/sweettest-baseline         gate: green on 0.6.1
        └── chore/holochain-0.7         gate: the above still green
            └── chore/holochain-0.7-js  gate: integration + acceptance + e2e green
```

Keeping the diffs based on `main-0.6` keeps each layer small and readable. If the stack is merged as a single unit at the end, it can target `sprout` directly instead.

The ordering matters: the Sweettest layer must be green on 0.6.1 **before** the 0.7 port lands. A suite written at the same time as the port proves nothing about regression. As a stack this stops being a convention I have to remember and becomes a merge requirement.

### Why Sweettest belongs in this line of work

Every test in this repo currently enters through the GraphQL layer: `tests/src` (2,996 lines, Tryorama), `clients/acceptance` (963 lines), `clients/playground-e2e` (144 lines). A `grep -rl "callZome\|call_zome"` across all of them returns nothing. There is no coverage at the zome boundary at all.

That is precisely the boundary Holochain 0.7 changes, and it is where the VF 1.0 validation rules added in this PR live. The `Proposal.purpose` immutability rule in 8d8f5720 is a good example: it is a validation-layer rule that is only exercised transitively today.

Also worth noting: Tryorama is no longer maintained by the Holochain team for 0.7 (holochain/holochain#5546). The community fork `@holochain-open-dev/tryorama@0.20.0` will keep the existing suites running, but Sweettest is the recommended path going forward and the repo has no Rust test crate today.

### Layer detail

1. **`test/sweettest-baseline`.** Add `tests/sweettest` as a native workspace member. The zome crates target wasm32, so `holochain` cannot be their dev-dependency; the test crate loads `workdir/hrea.dna` via `SweetDnaFile`. Starting scope is commitment, economic_event and proposal, covering create/update/delete, one link type, and the purpose immutability rule, then expanding.

2. **`chore/holochain-0.7`.** `hdi = "=0.8.0"`, `hdk = "=0.7.0"`, `holochain_serialized_bytes = "=0.0.57"`. Sweettest features become `encryption` and `wasmer-sys-cranelift`.

3. **`chore/holochain-0.7-js`.** holonix `main-0.7` and `nodejs_24`; `@holochain/hc-spin ^0.700.0`; `@holochain/client ^0.21.0` in `tests` and `ui`; Tryorama moved to the community fork with the local patch re-based; and in `modules/vf-graphql-holochain`, de-generic `SignedActionHashed` and move common action fields to `action.hashed.content.header.*`.

### Versioning and releases

Holochain 0.7 changes the DNA hash and provides no data migration path, so this is a minor bump rather than a patch.

| Artifact | Current (HC 0.6.1) | Through 0.7 | Stable |
|---|---|---|---|
| hApp tag | `happ-0.4.0-beta` | `happ-0.5.0-beta.1`, `.2`, ... | `happ-0.5.0` |
| npm module | `0.600.0-rc.0` | `0.700.0-rc.0`, ... | `0.700.0` |

`happ-0.4.0-beta` stays put as the Holochain 0.6 line for consumers that are not ready to move. Successive `-beta.N` prereleases will be cut from this line, the same way `happ-0.4.0-beta` was.

### How this gets validated before merge

The prereleases are the point, not a side effect. Both requests-and-offers and nondominium use hREA extensively, and both will run the `happ-0.5.0-beta.N` builds ahead of the stable tag. Two real hApps exercising the VF 1.0 API surface and the 0.7 runtime in practice is a stronger merge signal than review alone can give on a change this size, and any problems they surface get fixed in the beta line rather than after a stable release.

Feedback on the layer boundaries is welcome, particularly if you would prefer the split drawn somewhere else.

Soushi888:
> **SoushAI analysis.** Drafted by Soushi's AI assistant, reviewed and posted by @Soushi888.

## `happ-0.5.0-beta` is ready to cut

The stack has landed. #409, #410, #414 and #415 all merged into this branch today, the three feature branches are gone, and this is the only PR left open. Cut commit is `a2f8dc0f`, and CI is green on it.

**What that green means, measured on the cut commit rather than on a branch above it:**

- **Sweettest, at the zome boundary:** 47 tests across six binaries, 0 failed, against a live Holochain 0.7 conductor.
- **Acceptance battery, at the GraphQL adapter:** `71/71 steps green`, same conductor, on `@holochain/client` 0.21.

Holochain 0.7 works end to end: zomes on `hdk 0.7.0` / `hdi 0.8.0`, holonix `main-0.7`, the JS side on client 0.21 and `hc-spin` 0.700.

## Four things worth flagging to reviewers

**Tryorama is retired, and that is what unblocked 0.7.** The last published `@holochain/tryorama` is 0.19.2 and depends on `@holochain/client ^0.20.4`. No version targets 0.7 or client 0.21, so the old suite could not have run against this release under any arrangement. It was replaced at two layers rather than ported.

**Integrity validation is real now.** The zome previously validated only `EconomicEvent` and `Proposal`; everything else was a permissive stub. Every shared ValueFlows validator now has a rejection test asserting on that rule's own message, with an accepting case sitting on each boundary. Consumers should expect writes a `happ-0.4.0-beta` conductor accepted to be rejected here, with a message naming the rule.

**One known defect ships with it, filed rather than rushed:** #416. `update_rea_economic_event` declares eleven updatable fields and applies only `note`, returning success for the other ten. Pre-existing, not a regression from the 0.7 work, and fixing it means touching link-index maintenance on the most-linked entity in the system. That did not seem like work to do in the hours before a tag.

**Documentation is written against the merged state**, including a new `docs/consuming-a-release.md` for integrators: the two artifacts, composing `hrea.dna` into your own `happ.yaml`, the version pinning contract, the 0.6 to 0.7 breaking changes, and what does and does not carry over from `happ-0.4.0-beta`.

## A question for @LeosPrograms and @fosterlynn: should this branch merge to `sprout`?

Once `happ-0.5.0-beta` is tagged, this branch will have produced **two releases that `sprout` has never seen**.

`happ-0.4.0-beta` was cut from here at `e35e5483` on 2026-07-27 and is not an ancestor of `sprout`. `happ-0.5.0-beta` will be cut from `a2f8dc0f`, also here. Meanwhile `sprout` still points at `cefbdfb4`, and the count is one-directional: this branch is **62 commits ahead of `sprout`, and `sprout` is 0 ahead of it**. Merging is a clean fast-forward with nothing to reconcile.

Both releases are in real use rather than theoretical. [Requests and Offers](https://github.com/happenings-community/requests-and-offers) pins `happ-0.4.0-beta` and fetches `hrea.dna` from that release in its `postinstall`, and it is the project waiting on `0.5.0-beta` to make its own move to Holochain 0.7. Nondominium is in the same position.

The concern with leaving it as is: an integration branch that has outlived two releases is no longer an integration branch, it is the de facto trunk, and `sprout` is drifting into a fiction. Anyone who clones and builds `sprout` today gets neither the ValueFlows 1.0 surface nor Holochain 0.7.

So the question is simply whether you would like this merged, and if not, what would need to change first. Happy either way; it just seems worth deciding deliberately rather than by default.

## After it merges: issues and the project board

If the merge goes ahead, I would like to propose a reorganisation of the issues and the Project board to plan the next phases properly. The follow-ups are already identifiable from this branch, and several are naturally sequenced rather than parallel:

- the pending VF 1.0 deprecations that need @fosterlynn's call, since they are breaking: removing `inScopeOf` from `EconomicEvent` / `Intent` / `Proposal` / `Plan`, and `atLocation` from `EconomicEvent` / `Commitment` / `Intent`
- `BatchLotRecord`, replacing the removed `ProductBatch`, and the relational wiring that depends on it (`EconomicResource.lot`, and `.currentLocation` to `SpatialThing`)
- the legacy `pass` / `fail` actions, kept until their removal is agreed
- `QuantityValue.has_unit` referential validation, deferred because it adds a DHT read per quantity
- `agentRelationship` and its role types, declared by the base schema and unimplemented here
- #416, and the remaining sparse-update sweep it belongs to

I will put a concrete proposal together rather than just a list, once we know whether the merge is happening.

LeosPrograms (approved):
It looks generally ok to me. I added some test cases that weren't covered by the new tests, which all pass. Also removed package-lock.json.

Soushi888:
Thanks @LeosPrograms ! I merge and do smaller PRs next !

PR #418, opened 2026-09-17
ci(adapter): publish @valueflows/vf-graphql-holochain from CI

Soushi888:
> **SoushAI analysis.** Drafted by Soushi's AI assistant, reviewed and posted by @Soushi888.

## Context

`happ-0.5.0-beta.1` requires Holochain 0.7 and `@holochain/client ^0.21.0`. The adapter that talks to it, `0.700.0-rc.0`, exists only in this repository. The latest version on npm is `0.600.0-rc.0`, published 2025-12-05, which is the 0.6 line and depends on `@holochain/client ^0.20`.

So there is currently no install line an integrator can follow against the release we published. Publishing has been a manual act from a maintainer's own credentials (`yarn run publish:graphql:adapter`), and that is a large part of why it has not happened.

## What this adds

`.github/workflows/publish-adapter.yml`, triggered by tags matching `npm-modules-*` (the prefix this repo already used for `npm-modules-0.0.2-alpha.1`) and by `workflow_dispatch` for a publish without a tag.

It deliberately does not ride the `happ-*` tag. The adapter versions independently of the hApp, and `release.yml` spends the better part of an hour building WASM artifacts this job has no use for.

Three gates before anything reaches the registry:

1. **Version agreement.** On a tag-triggered run, `npm-modules-<v>` must match `modules/vf-graphql-holochain/package.json`, or the run fails naming both values.
2. **Schema surface.** `scripts/verify-purpose-schema.mjs` runs against the freshly built adapter, so a build that silently lost a ValueFlows 1.0 field cannot be published.
3. **No silent republish.** `npm view <name>@<version>` is checked first. npm answers a duplicate version with a 403, which reads like an auth failure and costs an hour of looking in the wrong place.

The workflow uses the same Nix and Cachix setup as `test.yml` and `release.yml`, so node stays pinned in `flake.nix` rather than being declared a second time in workflow YAML.

**`dist_tag` is an input, defaulting to `latest`.** npm marks a publish `latest` whether or not the version is a prerelease, which is how `0.600.0-rc.0` became the latest version while the 0.7 line was the one being recommended. Passing `next` is now a deliberate choice rather than something you remember after the fact.

`package.json` gains `repository`, `homepage` and `bugs`. npm refuses to attest provenance for a package that does not declare where its source lives, and the publish step passes `--provenance`.

## What it needs from a maintainer

One repository secret: **`NPM_TOKEN`**, a granular npm access token with write access to `@valueflows/vf-graphql-holochain`. Only a package owner can mint it; `npm owner ls` currently shows elfpavlik, pospi, mayel, sprillow-connor and leosprograms.

Until that secret exists the workflow is inert, and merging it changes nothing. That is the intended order: merge, add the secret, then run it once with `dry_run: true` to watch it pack without publishing.

## Verification

The parts that can be verified without the secret were:

- YAML parses, one job, ten steps.
- The version gate was exercised on both branches: `npm-modules-0.700.0-rc.0` against `0.700.0-rc.0` proceeds, `npm-modules-0.699.0` is rejected naming both values.
- The duplicate guard was exercised both ways against the live registry: `0.700.0-rc.0` is absent so a publish proceeds, `0.600.0-rc.0` exists so a republish is blocked.
- `package.json` still parses as JSON after the metadata addition.

The publish step itself cannot be proven until the token exists. The `dry_run` input is there for exactly that first run.

## Documentation

`docs/contributing.md` updated in the same commit:

- **Continuous integration** now describes three workflows rather than two, with what `publish-adapter.yml` checks and the secret it needs.
- **Publishing the GraphQL adapter** now leads with the CI path, keeps the manual `yarn run publish:graphql:adapter` as the documented fallback, and states the `latest` dist-tag behaviour that caught us on `0.600.0-rc.0`.

## One follow-up, deliberately left out

`finish-build.js` has its README copy commented out, so the published package shows no README on npm. Uncommenting it is one line, but it is not this PR's job.


Soushi888:
> **SoushAI analysis.** Drafted by Soushi's AI assistant, reviewed and posted by @Soushi888.

@leosprograms two asks here, one urgent and one structural. They are independent: the first unblocks people today, the second stops us needing the first.

## 1. Please publish `0.700.0-rc.0` by hand, now

This does not need the PR. Anyone trying to use `happ-0.5.0-beta.1` right now has nothing on the registry that can talk to it: the latest published adapter is `0.600.0-rc.0` from December, which is the 0.6 line and depends on `@holochain/client ^0.20`. The version in the tree is already correct, so it is one command from an owner's machine:

```bash
git checkout sprout && git pull
yarn install
yarn run publish:graphql:adapter
```

Two things to know before you run it:

- **The dist-tag.** npm marks a publish `latest` whether or not the version is a prerelease, which is how `0.600.0-rc.0` became latest while 0.7 was the line we were recommending. `publish:graphql:adapter` passes no tag, so this will take `latest`. I think that is right, since 0.7 is what we want a bare `npm install` to get. If you disagree, publish from `modules/vf-graphql-holochain/build/` with `npm publish --access=public --tag next` instead.

- **Order against this PR.** Published from `sprout` today, the package carries no `repository` field, so npm shows no source link. If you merge this PR first and publish from the merged branch, it gets one. Either is fine; the metadata is cosmetic for a manual publish, and provenance only exists on the CI path anyway.

If you would rather not, say so and I will look at what it takes to get an owner to do it another way. But it is currently the one thing standing between the release and anyone using it.

## 2. The token, so this stops being a manual act

This PR is inert until there is an `NPM_TOKEN` repository secret: a **granular access token**, scoped to `@valueflows/vf-graphql-holochain` only, read and write, with an expiry you are comfortable with. Only a package owner can mint it, and `npm owner ls` says that is you, @elf-pavlik, @pospi, @mayel and @sprillow-connor.

The sequence I would suggest:

1. Merge this PR. Nothing happens.
2. Mint the token, add it under Settings, Secrets and variables, Actions, as `NPM_TOKEN`.
3. Run the workflow once from the Actions tab with `dry_run: true`. It builds, verifies the schema surface, and packs without publishing, so we see the tarball before anything is public.
4. From then on, publishing is `git tag npm-modules-<version> && git push origin npm-modules-<version>`.

Separately, and only if you are comfortable with it: adding me as an owner (`npm owner add <user> @valueflows/vf-graphql-holochain`, or membership in the `@valueflows` org) would mean the manual fallback does not always route through you. Your call entirely, and the CI path makes it much less necessary.

Happy to walk through any of this synchronously if that is faster.

